PR #1, opened 2026-08-20
fix(compiler): lower TypeScript enums, keep JSON __proto__ own, size compile stacks

williamhogman:
Three ways the per-file pipeline could fail a whole command instead of one file.
Each was found by an adverse or property test, and each test is in the PR.

**A TypeScript `enum` panicked the compiler.** oxc's TS transform needs scoping
built with `SemanticBuilder::with_enum_eval(true)` to lower an enum, and panics
without it. So this took `oj` down, in dev and in build:

```ts
export enum Direction { Up, Down }
```

Same for a bundle-mode factory and for an `oj.config.ts` that declares one.
Enabled on all three transform paths.

**A `__proto__` key in a JSON import set the prototype instead of a property.**
`json::to_esm` splices the document straight into an object literal, where
`{"__proto__": x}` is special — so `import data from './d.json'` disagreed with
`JSON.parse` on the same bytes, and `data.polluted` came back `true` for
`{"__proto__":{"polluted":true}}`. In the factory path the `__proto__` getter
replaced the namespace object's prototype outright. Documents containing the key
now go through `JSON.parse`, and it is never emitted as a named export.

**Deep nesting aborted the process.** Parsing, transforming and printing each
recurse once per bracket level, and a stack overflow aborts rather than failing
the file — a generated or minified module could take the dev server with it. It
went over below 1000 levels on the 2 MiB a runtime thread gets by default. The
CLI now installs `oj_compiler::COMPILE_STACK_SIZE` as the runtime's thread stack
size, which is also the size the new tests exercise. There is no depth that is
safe for every stack; this makes the envelope explicit and far above anything
hand-written.

### Tests

`crates/oj_compiler/tests/`, 64 tests in three suites:

- **adverse** — malformed and hostile source (unterminated everything, deep
  nesting, unicode and bidi identifiers, lone surrogates), hostile import
  specifiers, a rewriter that returns junk, JSON that is not JSON, `__proto__`
  keys, every `require` shape, glob and dynamic-import patterns that try to
  escape.
- **edge_cases** — empty files, BOMs, shebangs, CRLF, TypeScript-only syntax
  including `enum`, JSX per extension, top-level await, every export form,
  dev-vs-prod instrumentation, sourcemap omission for synthesized modules.
- **properties** — totality over arbitrary bytes, output validity by re-parsing
  what was emitted, determinism, output as a parser fixed point, import order
  and one-shot rewriting, mode-independence of the module graph.

`cargo test --workspace` and `node e2e/run.mjs` (both modes) pass, as does the
`playground` production build with every CI assertion.
---

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -266,6 +266,21 @@ dependencies = [
 ]
 
 [[package]]
+name = "bit-set"
+version = "0.8.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "08807e080ed7f9d5433fa9b275196cfc35414f66a0c79d864dc51a0d825231a3"
+dependencies = [
+ "bit-vec",
+]
+
+[[package]]
+name = "bit-vec"
+version = "0.8.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5e764a1d40d510daf35e07be9eb06e75770908c27d411ee6c92109c9840eaaf7"
+
+[[package]]
 name = "bitflags"
 version = "2.13.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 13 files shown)


raphamorim:
Merged to main manually in 035c2a0 (the compiler fixes and tests) and ba8ee03 (the temp-directory test fix), rebased on top of current main with the messages tidied up. Thanks @williamhogman, great catches on the enum lowering, JSON __proto__ handling, and the compile stack size.

PR #2, opened 2026-08-20
fix(env): keep non-ASCII dotenv values intact

williamhogman:
`expand` walked the value byte by byte and pushed each byte as a `char`, so every
multi-byte character in an expanded value came out mojibaked:

```sh
# .env
VITE_TITLE=café
```

reached the app as `cafÃ©`. Single-quoted values escaped it only because they
skip expansion entirely, which is why this hid for a while: `'café'` was fine and
`café` was not. Fixed by copying whole characters.

Also drops a dead index. The unclosed-`\${` branch computed a `next` position
nothing could read; mutation testing flagged it as an unkillable mutant, which is
the tell. The branch now says what it means with `Option`.

### Tests

`crates/oj_env/tests/`, 31 tests in two suites:

- **adverse** — malformed lines, self-referential and mutual expansion,
  unterminated quotes and braces, control bytes and CRLF, non-ASCII values byte
  for byte, `%`-soup in HTML, and secret confinement: near-miss prefixes
  (`vite_`, ` VITE_`, `XVITE_`), and a value that tries to break out of the JSON
  literal it is embedded in.
- **properties** — parsing is total over arbitrary bytes; single-quoted and
  metacharacter-free values are byte-exact; the individual `import.meta.env.*`
  defines and the aggregate object always have the same keys; no unprefixed
  variable is ever exposed, by name or by value.

One documented divergence is pinned rather than silently carried: Vite's
`/%(\S+?)%/g` leaves `%%KEY%%` untouched, where oj rescans from the inner `%` and
substitutes. Only adjacent percents differ.
---

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -266,6 +266,21 @@ dependencies = [
 ]
 
 [[package]]
+name = "bit-set"
+version = "0.8.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "08807e080ed7f9d5433fa9b275196cfc35414f66a0c79d864dc51a0d825231a3"
+dependencies = [
+ "bit-vec",
+]
+
+[[package]]
+name = "bit-vec"
+version = "0.8.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5e764a1d40d510daf35e07be9eb06e75770908c27d411ee6c92109c9840eaaf7"
+
+[[package]]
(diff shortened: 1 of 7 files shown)


raphamorim:
Merged to main in 0b10392, rebased onto current main (the temp-directory test fix was already landed, so only the env change and its tests came across) with the message tidied. Thanks @williamhogman, the byte-vs-char mojibake fix and the property tests are exactly right.

PR #3, opened 2026-08-20
fix(graph): walk HMR boundaries iteratively

williamhogman:
`climb` recursed once per level of the importer chain, so a change deep in a long
chain overflowed the stack of the runtime worker handling the file event — and a
stack overflow aborts the process, taking the dev server down rather than failing
one update. It went over at a few thousand levels on the 2 MiB a worker gets,
which barrel files and generated route trees reach.

Same traversal, explicit stack. The gray/black colouring, the visit order, the
"circular import involving X" report and the "reached entry X with no accepting
boundary" report are all unchanged. One behaviour change: a module missing from
the map now falls back to a full reload instead of indexing into it, since a full
reload is always a correct answer and a panic never is.

### Tests

`crates/oj_graph/tests/`, 17 tests in two suites:

- **adverse** — 200k-deep chains, 20k-wide fan-in, cycles of every length from 2
  up, diamonds, self-imports, and repeated re-linking that must leave no stale
  edges. Each runs on a thread with the 2 MiB a runtime worker gets, so a
  recursive walk shows up here rather than in someone's dev server.
- **model** — a model-based property test: random link/accept/change sequences
  applied to both the graph and a from-scratch reference model, comparing
  structure and every HMR decision after each step. The reference recomputes
  everything each time on purpose, so incremental bookkeeping drifting away from
  the truth is what it catches.

The checked-in proptest seed is not decoration: it is the case that catches a
walk which forgets to reset a module from gray to black after visiting its
importers, which turns a diamond into a reported cycle. I verified it fails
against that mutation.
---

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -266,6 +266,21 @@ dependencies = [
 ]
 
 [[package]]
+name = "bit-set"
+version = "0.8.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "08807e080ed7f9d5433fa9b275196cfc35414f66a0c79d864dc51a0d825231a3"
+dependencies = [
+ "bit-vec",
+]
+
+[[package]]
+name = "bit-vec"
+version = "0.8.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5e764a1d40d510daf35e07be9eb06e75770908c27d411ee6c92109c9840eaaf7"
+
+[[package]]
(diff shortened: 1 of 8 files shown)


raphamorim:
Merged to main in 7281785, rebased onto current main (the temp-directory test fix was already landed, so only the graph change and its tests came across) with the message tidied. Thanks @williamhogman, converting the boundary climb to an explicit stack is the right call, and the model-based test is excellent.

PR #4, opened 2026-08-20
fix(cache): validate keys and publish entries atomically

williamhogman:
Two things about how entries reach and leave the disk.

**A key that was not a digest steered the path.** `path_for` built a path out of
whatever string it was handed, so `cache.put("../../../../tmp/x", …)` wrote
outside the cache directory — my test wrote `/tmp/oj-escape.json` before the fix.
Every key in the tree comes from `key()` and is therefore a 64-character blake3
digest, so anything else is a caller bug and is now refused rather than resolved.
Latent today, but a cache is exactly the component someone later hands a
URL-derived key.

**Publication was not atomic under contention.** The temp file a `put` writes
through was named after the entry, so two builds sharing one `.oj-cache` — CI, an
agent fleet, the workloads oj exists for — could have one writer's rename publish
the file while another was still writing into it. A reader then sees a short entry
and evicts a good one. Each writer now gets its own temp name, so the rename is
the only way an entry becomes visible, and a failed write cleans up after itself.

### Tests

`crates/oj_cache/tests/`, 24 tests in three suites plus a shared harness:

- **faults** — garbage, truncated, wrong-type and wrong-schema entries; an entry
  written by an older schema (fields must default, not fail); unknown fields from
  a newer oj; a read-only cache dir and a read-only shard; a cache dir that is a
  file; keys that are not digests; version isolation. Every one has the same
  correct outcome: a miss, and a cache that still works afterwards.
- **concurrency** — 8 writers on one key publish a complete entry; readers racing
  writers see the exact entry or nothing, never a torn one; 400 distinct keys all
  land with one file each; corruption by a crashed peer races safely with
  rewriting.
- **properties** — keys are hex digests, deterministic, independent of the cache
  directory, and injective over (source, url, mode, version); entries round trip
  exactly, including `\u{2028}`, NUL and emoji; non-digest keys never write.
---

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -266,6 +266,21 @@ dependencies = [
 ]
 
 [[package]]
+name = "bit-set"
+version = "0.8.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "08807e080ed7f9d5433fa9b275196cfc35414f66a0c79d864dc51a0d825231a3"
+dependencies = [
+ "bit-vec",
+]
+
+[[package]]
+name = "bit-vec"
+version = "0.8.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5e764a1d40d510daf35e07be9eb06e75770908c27d411ee6c92109c9840eaaf7"
+
+[[package]]
 name = "bitflags"
 version = "2.13.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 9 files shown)


raphamorim:
Merged to main in b91d21f, rebased onto current main (the temp-directory test fix was already landed, so only the cache change and its tests came across) with the message tidied. Thanks @williamhogman, the key validation closes a real path-traversal hole and the per-writer atomic publish is exactly what the shared-cache scenario needs.

PR #5, opened 2026-08-20
fix(css): classify modules by filename, and pin the target matrix

williamhogman:
`is_css_module` matched `.module.` anywhere in the last path segment, query
included, so `styles.css?x=.module.` was compiled as a CSS module — scoped class
names and an exports map for a plain stylesheet. It now strips the query first,
the way `is_sass` next to it already did.

The other half of this PR is two behaviours that had no test at all, which
mutation testing surfaced (every mutant of them survived):

- **the browser target matrix** — it decides what is downlevelled for whom, and a
  matrix that decoded to version 0 would downlevel everything. Asserted through
  behaviour: `clamp()`, modern color syntax and `aspect-ratio` survive because
  the configured versions support them, while nesting and logical properties are
  lowered because the oldest target does not.
- **the CSS Modules naming pattern** — the scoped name shows up in devtools, in
  snapshots and in the exports map. Also pinned: the hash comes from the path,
  not the contents, so editing a stylesheet does not rename its classes.

### Tests

`crates/oj_css/tests/adverse.rs` (17 tests) plus the two unit tests above:
malformed and truncated stylesheets, 20k-rule and 1MB-value files, nesting within
the supported envelope, module scoping collisions across files, Sass error
reporting, import confinement with and without a load path, and minification as
a formatting choice only (re-parsing either form gives the same output).

One case is `#[ignore]`d rather than pretended away: unbounded Sass recursion
(`@mixin a { @include a; }`) aborts the process. `grass` has no call-depth limit
and Rust cannot catch a stack overflow, so containing it needs process isolation,
not a bigger stack. The test stays runnable under `--ignored` so the behaviour can
be re-checked after a `grass` upgrade.
---

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -266,6 +266,21 @@ dependencies = [
 ]
 
 [[package]]
+name = "bit-set"
+version = "0.8.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "08807e080ed7f9d5433fa9b275196cfc35414f66a0c79d864dc51a0d825231a3"
+dependencies = [
+ "bit-vec",
+]
+
+[[package]]
+name = "bit-vec"
+version = "0.8.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5e764a1d40d510daf35e07be9eb06e75770908c27d411ee6c92109c9840eaaf7"
+
+[[package]]
 name = "bitflags"
 version = "2.13.1"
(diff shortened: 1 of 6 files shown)


raphamorim:
Merged to main in a6bf9e7, rebased onto current main (the temp-directory test fix was already landed, so only the css change and its tests came across) with the message tidied. Thanks @williamhogman, stripping the query before the .module. check matches Vite's cssModuleRE, and pinning the target matrix and scoped-name pattern by behaviour is the right guard.

PR #6, opened 2026-08-20
fix(config): bound evaluation and check the shape of what it returns

williamhogman:
A config file is code oj runs, and it ran unbounded.

```ts
// oj.config.ts
while (true) {}
export default {};
```

`oj dev` hung forever, no output, no timeout. An allocation loop grew until the OS
stepped in. The QuickJS runtime now gets a 5s deadline through an interrupt
handler and a 64 MiB heap cap, and a config that hits the deadline says so
instead of surfacing something from the engine's internals.

Second half: every field of `OjConfig` is optional, so serde read a JSON array as
a struct of defaults — `export default []` was silently accepted as a valid empty
config, which is a config that looks like it worked. And a file that exported
nothing failed with `invalid type: null, expected struct OjConfig`, which tells
the user nothing about what to do. The evaluated value is now checked to be an
object first, and the message names the actual problem
("no config was exported; a config file must `export default` an object").

### Tests

`crates/oj_config/tests/adverse.rs`, 25 tests: a config that never finishes,
allocates without bound, recurses, throws, has a throwing getter, is cyclic,
exports nothing, or is not an object; fields of the wrong type; unknown fields
ignored rather than fatal; TypeScript stripping; the `defineConfig` import being
stripped while the sandbox still provides it; how far the sandbox reaches
(`require`, `fetch`, `process.binding` are all undefined, and `process` has
exactly one key); candidate filename precedence; a config directory instead of a
file; invalid UTF-8; and that loading twice gives the same answer.

The 5s bound is what three of those tests wait on, so the suite takes ~5s wall
clock — they run in parallel, so it is 5s total rather than 15.
---

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -266,6 +266,21 @@ dependencies = [
 ]
 
 [[package]]
+name = "bit-set"
+version = "0.8.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "08807e080ed7f9d5433fa9b275196cfc35414f66a0c79d864dc51a0d825231a3"
+dependencies = [
+ "bit-vec",
+]
+
+[[package]]
+name = "bit-vec"
+version = "0.8.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5e764a1d40d510daf35e07be9eb06e75770908c27d411ee6c92109c9840eaaf7"
+
+[[package]]
 name = "bitflags"
 version = "2.13.1"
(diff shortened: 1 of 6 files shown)


raphamorim:
Merged to main in 0eaba64, 3-way merged onto current main so it composes with the earlier enum-in-config fix (the temp-directory test fix was already landed, so only the config change and its tests came across), message tidied. Thanks @williamhogman, bounding the QuickJS eval with a deadline and heap cap and rejecting a non-object config are both exactly right for a sandboxed config.

PR #7, opened 2026-08-20
fix(server): decode request paths, key the dep optimizer per list, name the sidecars

williamhogman:
Four bugs on the request path and around the node sidecars.

**Request paths were never percent-decoded.** A component named `Cool Button.tsx`
arrives as `Cool%20Button.tsx`; a stylesheet named `café.css` arrives as
`caf%C3%A9.css`. Both 404'd in `oj dev` **and** `oj preview` — files the app
imports, unreachable. Decoding now happens before the path is treated as a path,
which also puts it before the traversal guard, so `%2e%2e` is caught by the same
check as `..` instead of passing through as a literal segment. `/@fs` decodes too
and stays allow-listed.

**The dep optimizer's cache key could not tell its lists apart.** `include`,
`exclude`, `entries` and `dedupe` were folded into one flat sequence, so moving a
package from `include` to `exclude` left the key unchanged and the stale prebundle
in place — excluding a dep did nothing until something else invalidated the cache.
Each list is now tagged.

**All three sidecars said "tailwind".** A Less file that failed to compile
reported `tailwind sidecar died (is tailwindcss installed?)`. Each sidecar now
names itself and the package to install. A request that times out also drops its
pending entry, which used to leak a waiter per timeout.

**`is_tailwind_css` was wrong in both directions.** It matched `@theme` anywhere,
so a comment mentioning `@theme` or a rule with `content: "@theme"` routed a plain
stylesheet through Tailwind and failed the build. Meanwhile
`@import "tailwindcss/preflight" layer(base);` — the documented v4 form — was not
matched at all, which silently drops every utility class from the page. Directives
now count only where a directive can start, and the import matches the package
rather than an exact string (`tailwindcss/theme` yes, `tailwindcss-animate` no).

### Tests

`e2e/awkward-paths.mjs` drives a real dev server and a real preview server: files
whose names need encoding are served, and seven spellings of a traversal (raw,
`%2e%2e`, `%2E%2E%2F`, `/@fs` with and without encoding) never return a file from
outside the app. I checked it fails against the unfixed code. Added to CI.

Unit tests for percent-decoding (including malformed escapes and double-encoding),
traversal rejection after decoding, filenames that need encoding, `html_entry_src`,
gate relevance, query classification, the optimizer's key and its manifest
handling, and 5 tests over the Tailwind detection in both directions.
---

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -54,6 +54,9 @@ jobs:
           node e2e/memory-cache.mjs
           node e2e/mtime-fastpath.mjs
           node e2e/hmr-hook-skip.mjs
+      - name: request paths (encoded filenames, traversal containment)
+        if: matrix.mode == 'unbundled'
+        run: node e2e/awkward-paths.mjs
 
   # i need to fix this or automate
   build:
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1900,6 +1900,7 @@ dependencies = [
  "reqwest",
  "serde_json",
  "simdutf8",
+ "tempfile",
  "tokio",
 ]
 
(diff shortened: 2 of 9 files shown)


raphamorim:
Merged to main in 5d118e0, squashed to a single commit and rebased onto current main (the temp-directory test fix was already landed, so only the server change, its tests, and the CI step came across), message tidied. Thanks @williamhogman, decoding before the traversal check, tagging each optimizer list in the key, and the per-kind sidecar errors with the fixed waiter leak are all solid.

PR #8, opened 2026-08-20
test(resolver): adverse suite over specifiers and package metadata

williamhogman:
No behaviour change. This pins what resolution does with input it did not write,
including the places where the answer is deliberately permissive — which is the
part worth having in writing.

24 tests in `crates/oj_resolver/tests/adverse.rs`:

- Degenerate specifiers (empty, `.`, `..`, `/`, NUL, 64 `../` segments, 4KB of
  one character) fail with a described error naming both the specifier and the
  importer.
- Malformed `package.json` and `tsconfig.json` never panic. JSONC in a tsconfig
  (comments, trailing commas) is understood, since real ones have both.
- An exports map keeps hiding what it hides, from every spelling of the path
  (`pkg/private`, `pkg/./private.js`, `pkg/../pkg/private.js`,
  `pkg/package.json`), and cannot be used to escape its own package.
- A `#import` stays private to its package; a self-reference honours the map.
- Symlinked deps resolve; symlink cycles terminate.
- Dedupe applies to bare specifiers only, matches scoped packages by their full
  name, falls back to the importer when the root has no copy, and is inert for a
  degenerate list. A package that merely starts with a deduped name is not
  deduped.
- Extension probing order is stable when every candidate exists at once.
- Resolution is deterministic and creates nothing on disk.

Two of these record a boundary rather than a requirement, so it cannot be quietly
changed later:

- **Resolution is not confinement.** A linked package or a monorepo sibling lives
  outside the project root, so paths outside it resolve — including through a
  `file://` specifier, which is the one URL scheme the Node algorithm treats as a
  path. The dev server's `/@fs` allow-list is what decides whether such a path may
  be *served*, and `e2e/awkward-paths.mjs` in #7 tests that half.
- **A broken `tsconfig.json` fails every resolution**, with a message naming the
  tsconfig — rather than guessing at an alias table and resolving imports to the
  wrong files.
---

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1879,6 +1879,7 @@ name = "oj_resolver"
 version = "0.0.45"
 dependencies = [
  "oxc_resolver",
+ "tempfile",
  "thiserror",
 ]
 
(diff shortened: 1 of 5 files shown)


raphamorim:
Merged to main in 4853d0d, squashed to a single commit and rebased onto current main (the temp-directory test fix was already landed, so only the resolver test suite came across). Thanks @williamhogman, good adverse coverage over specifiers and package metadata.

PR #9, opened 2026-08-20
docs: testing strategy, fuzz targets, mutation config, nightly CI

williamhogman:
Writes down what the other eight PRs were found by, so the method outlives the
bugs.

**`docs/development/testing.md`** maps the layers and lists what each suite
proves. The organizing idea is that suites are named after *how a thing can go
wrong* rather than after the module they cover — adverse input, boundary shapes,
contention, injected faults, properties — with five invariants doing most of the
work: totality (arbitrary bytes are a legal file), output validity (assert by
re-parsing, not by matching strings), confinement (no unprefixed variable reaches
the client, no request path reaches outside the project), determinism, and
degradation (the cache is an optimization; every failure costs a recompile and
nothing else).

Its **Deliberate boundaries** section is the part I would read twice: six
behaviours that look like gaps and are decisions, each pinned by a test —
recursion depth as a stack-size contract, unbounded Sass recursion aborting the
process (`grass` has no depth limit and Rust cannot catch a stack overflow),
early errors left for the browser to report, resolution deliberately not being
confinement, a broken tsconfig failing loudly, and one documented divergence from
Vite's `%%KEY%%` handling.

**`fuzz/`** adds five cargo-fuzz targets over the untrusted-input surfaces. Each
asserts an invariant rather than merely "does not crash": compiled output
re-parses and is deterministic; the dotenv defines never leak an unprefixed
variable; valid JSON always yields a valid module with one default export and no
bare `__proto__` key; `%KEY%` substitution only touches known keys and never
rescans its own output; any cache entry that deserializes round trips unchanged.
`fuzz_compile` skips inputs nested past 200 brackets, since depth is the
stack-size contract above and without the filter the fuzzer spends all its time
rediscovering that parsers recurse.

**CI** keeps the fuzz targets compiling on every PR (they are outside the
workspace, so they would rot silently). A nightly workflow runs each for ten
minutes, uploads crash artifacts, and runs the `#[ignore]`d tier that aborts on
purpose.

**`.cargo/mutants.toml`** scopes cargo-mutants to the pure crates — the dev
server and CLI are covered from `e2e/`, which it cannot drive. It is what
surfaced the untested CSS target matrix and module naming pattern (#5) and one
dead expression (#2). `oj_env` and `oj_css` now come back 95 caught / 6 timed out
/ 6 missed of 108, and all six survivors are equivalent mutants — documented, so
a new survivor there means a real gap rather than noise.

Caveat worth stating: the fuzz targets are compile-checked, not run. cargo-fuzz
needs a nightly toolchain, which the machine I wrote this on does not have; the
nightly workflow is where they would first actually execute.
---

### DIFF
--- a/.cargo/mutants.toml
+++ b/.cargo/mutants.toml
@@ -1,0 +1,14 @@
+# cargo-mutants configuration. Informational, run on demand rather than in CI:
+# it is a check on the suites, not a gate on a change.
+#
+# Scoped to the pure, correctness-critical crates. The dev server and the CLI
+# are covered end to end from `e2e/`, which cargo-mutants cannot drive.
+examine_globs = [
+    "crates/oj_env/src/**",
+    "crates/oj_graph/src/**",
+    "crates/oj_cache/src/**",
+    "crates/oj_css/src/**",
+    "crates/oj_compiler/src/json.rs",
+]
+timeout_multiplier = 4.0
+minimum_test_timeout = 60
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,6 +16,10 @@ jobs:
       - uses: actions/checkout@v4
       - uses: Swatinem/rust-cache@v2
       - run: cargo test --workspace
+      # The fuzz crate is outside the workspace; keep its targets compiling so
+      # they cannot rot between nightly fuzzing runs.
+      - name: fuzz targets still build
+        run: cargo check --manifest-path fuzz/Cargo.toml --all-targets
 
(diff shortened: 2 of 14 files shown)


raphamorim:
Merged to main in a5c401f, squashed to a single commit and rebased onto current main (the temp-directory test fix was already landed; the CI change was 3-way merged so it composes with the awkward-paths step). Thanks @williamhogman, the fuzz targets, mutation config, and nightly workflow are a great addition to the testing strategy.

PR #10, opened 2026-08-20
fix(build): inline SVGs with CRLF endings, and give every asset a mime type

williamhogman:
Three bugs in how the production build emits assets.

**An inlined SVG with CRLF line endings failed the build.**

```
Error: build failed: PARSE_ERROR "Unterminated string"
```

`emit_or_inline` stripped `\n` from the SVG text and interpolated the result into
a string literal, which leaves the `\r` behind — and a raw carriage return cannot
appear in a JavaScript string literal at all. So a perfectly valid SVG, checked
out by git on Windows or anywhere with `core.autocrlf`, broke the build with a
parse error pointing at generated code the user never wrote. I reproduced it with
a two-file app before fixing it.

The data url now drops every line terminator (including U+2028 and U+2029, which
are string-literal terminators too) and percent-encodes the backslash. And the
module is emitted through a JSON-encoded literal, so no asset url — derived, as
they all are, from a file oj did not write — can produce something that does not
parse.

**`bmp` and `mov` inlined as `application/octet-stream`.** `is_build_asset`
accepts both, `asset_mime` had no arm for either, so an inlined BMP became a data
url the page will not render. Added, along with a test that walks every extension
`is_build_asset` accepts and asserts each has a real type — the two tables are
supposed to agree, and nothing was checking that.

**`content_hash` used `DefaultHasher`.** The standard library explicitly makes no
promise about that hasher's output across releases, and these hashes become asset
filenames that end up in caching headers. A deploy's cache keys should not change
because oj was rebuilt with a newer Rust. Switched to blake3, which the workspace
already depends on.

### Tests

Unit tests around the two extracted helpers (`svg_data_url`, `export_default_url`)
that assert by re-parsing what they emit, base64 against the RFC 4648 vectors,
the asset mime and classification tables against each other, content hashes
pinned by value, `transitive_imports` against cycles and self-imports, and asset
name sanitization.

`e2e/assets-inline.mjs` grows a CRLF SVG and a BMP: the SVG must inline with no
line terminator in the data url and its content intact, and the BMP must inline
as `data:image/bmp;base64,`. It fails against the unfixed build — I checked by
stashing only the Rust change.

`cargo test --workspace`, `node e2e/run.mjs`, the eight build-output e2e scripts
(manifest, assets, css-code-split, modulepreload, manual-chunks, css-rebase,
query-assets, assets-inline) and the `playground` production build all pass.

---

together cleanly — I verified both. Each also carries the same two-line `tmp()`
fix in `crates/oj/src/start_dev.rs`, because without it

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1792,6 +1792,7 @@ dependencies = [
  "anyhow",
  "arcstr",
  "axum",
+ "blake3",
  "clap",
  "notify",
  "oj_compiler",
--- a/crates/oj/Cargo.toml
+++ b/crates/oj/Cargo.toml
@@ -23,6 +23,7 @@ rolldown_plugin = "1.2.5"
 rolldown_utils = "1.2.5"
 arcstr = "1.2.0"
 serde_json.workspace = true
+blake3.workspace = true
 oj_css.workspace = true
 axum.workspace = true
 notify.workspace = true
--- a/crates/oj/src/build.rs
+++ b/crates/oj/src/build.rs
@@ -183,8 +183,10 @@ fn asset_mime(ext: &str) -> &'static str {
         "ttf" => "font/ttf",
         "otf" => "font/otf",
         "eot" => "application/vnd.ms-fontobject",
+        "bmp" => "image/bmp",
         "mp4" => "video/mp4",
         "webm" => "video/webm",
+        "mov" => "video/quicktime",
         "mp3" => "audio/mpeg",
         "wav" => "audio/wav",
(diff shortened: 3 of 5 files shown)


raphamorim:
Merged to main in 84c1586, squashed to a single commit and rebased onto current main (the temp-directory test fix was already landed). Thanks @williamhogman, dropping every line terminator from inlined SVGs, JSON-encoding the url module, and pinning content_hash to blake3 so filenames do not drift with the toolchain are all right.

PR #11, opened 2026-08-20
fix(ssr): confine the module runner's reads and server-function calls

williamhogman:
Two endpoints that `oj dev --ssr` adds take a path straight from the request.

**`GET /@ssr-module?id=<path>`** read whatever `id` named:

```
$ curl 'http://127.0.0.1:5199/@ssr-module?id=%2FUsers%2Fme%2F.claude%2Fnotes.mjs'
200  <the file's source>
```

Any compilable file on the machine came back with its contents in the body.

**`POST /__oj_fn`** joined the body's `module` field onto the project root
without normalising it, so `../../../../x.server.mjs` resolved outside the
project — and the module runner imported it and called the named export:

```
$ curl -X POST http://127.0.0.1:5199/__oj_fn -H 'content-type: text/plain' \
    -d '{"module":"../../../../tmp/outside.server.mjs","name":"leak","args":[]}'
200 "PWNED"
```

That `content-type: text/plain` matters: a JSON body sent with it is a CORS
*simple* request, so a browser sends it cross-origin with no preflight. Any page
a developer had open could import and execute a module from their filesystem,
and anything on the LAN could when the server was started with `--host`. I
confirmed both against a real dev server, before and after.

Neither affects the production server. `__ojDispatch` there resolves against a
table globbed from `**/*.server.*` at build time, which is a closed set. The fix
gives dev the same bound:

- a server-function call must name an existing `*.server.*` file that stays
  inside the project after normalisation;
- the module endpoint serves the project, anything under a `node_modules` (a
  dependency is not a secret, and it is the path every real SSR import takes),
  and whatever `server.fs.allow` names.

Paths that are not real files pass through untouched — a virtual id is not a
filesystem read, and only the plugin host can resolve one, to what a plugin
generated. That keeps `virtual:` modules and plugin-provided ids working, which
is what the first version of this patch broke and the existing SSR e2e caught.
Canonicalisation means a symlink planted inside the project does not widen it
either.

### Tests

Unit tests over both policies — the project/dependency/allow-list cases, the
denied spellings, the virtual-id pass-through, and a symlink escape.

`e2e/ssr-dev.mjs` now drives both endpoints against a real dev server: a module
outside the project, a relative traversal, an ordinary non-server module, and
`/etc/passwd`, all of which must be refused, while a genuine server function,
project files and dependencies must still work. It reports `200 "PWNED"` against
the unfixed code — I checked by stashing only the Rust changes.

---

together cleanly — I verified both. Each also carries the same two-line `tmp()`
fix in `crates/oj/src/start_dev.rs`, because without it

### DIFF
--- a/crates/oj/src/ssr_dev.rs
+++ b/crates/oj/src/ssr_dev.rs
@@ -80,6 +80,65 @@ pub async fn ssr_dev(
     Ok(())
 }
 
+/// Resolve the `module` field of a server-function call to a file oj is willing
+/// to import.
+///
+/// The field comes off the request body, so it is attacker-controlled -- and a
+/// browser can POST here cross-origin without a preflight, since a JSON body
+/// with `content-type: text/plain` is a simple request. Unchecked, that is
+/// "import and run any module on the developer's machine" from any page they
+/// have open. Two conditions, matching what the production dispatch table
+/// allows: inside the project, and named like a server module.
+fn server_fn_module(root: &Path, module_url: &str) -> Option<PathBuf> {
+    const SERVER_SUFFIXES: [&str; 8] = [
+        ".server.ts",
+        ".server.tsx",
+        ".server.js",
+        ".server.jsx",
+        ".server.mts",
+        ".server.mjs",
+        ".server.cts",
+        ".server.cjs",
+    ];
+
+    let rel = module_url.split('?').next().unwrap_or(module_url);
+    let rel = rel.trim_start_matches('/');
+    if rel.is_empty() {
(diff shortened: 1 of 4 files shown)


raphamorim:
Landed on `main` as 5141619 (squashed, authored by @williamhogman). Verified the confinement against a real `oj dev --ssr`: the module and `__oj_fn` endpoints now reject traversal / out-of-project paths, while genuine server functions, project files and dependencies still work, matching Vite's `server.fs.allow` model. Thanks William!

PR #13, opened 2026-08-21
add rolldown hook filters to the start client-bundle plugins

aeriksson-lovable:
Part 2 of 14 in a series. Depends on #12 — this PR's diff includes the commits of the PRs below it in the series; to see only this part: https://github.com/aeriksson-lovable/oj/compare/pr/01-orphan-hardening...pr/02-rolldown-hook-filters. Series overview in #12.

---

## Problem

oj bundles the browser entry for a TanStack Start app with rolldown. The
bundler runs in a Node.js child process. The script is
`crates/oj_server/src/assets/start/bundle-client.mjs`. oj spawns it from
`bundle_client_entry` in `crates/oj/src/start_dev.rs` on every boot and on
every watch rebuild.

Four dev plugins run inside this bundle:

- `server-fn-client`, defined in
  `crates/oj_server/src/assets/start/bundle-client.mjs`. It rewrites
  `createServerFn` calls to client RPC stubs and applies user-plugin
  transforms to user TypeScript files.
- `oj-assets`, `oj-vite-plugins`, and `node-builtin-shims`, defined in
  `crates/oj_server/src/assets/start/rolldown-assets.mjs`.

Every hook on these plugins was a bare JS function. Rolldown runs its module
graph in native code. For a bare JS hook, rolldown must call into JS for
every candidate. Each call crosses the native-to-JS boundary. `load` and
`transform` hooks run once per module. `resolveId` hooks run once per import
edge.

approximately 18,600 client modules and a 788-line Vite config that
registers 53 plugins. Hardware: macOS arm64 (M-series), release build of oj.
On this app the client graph has approximately 50,000–100,000 import edges.
Almost every hook call failed an in-handler guard and returned null. For
example, the `server-fn-client` transform bails for `node_modules` files,
for virtual ids, and for files that are not `.ts`/`.tsx`. The bundle stage
took 4.4–5.7 s per run (warm `.oj-cache`, measurements serialized).

## Change

Rolldown accepts an object form for these hooks: `{ filter, handler }`. The
filter is a declarative pattern on the module id. Rolldown evaluates the
filter on the native side. Rolldown calls the JS handler only when the
filter matches. Non-matching modules never cross the native-to-JS boundary.

The change, step by step:

1. Convert every `resolveId`, `load`, and `transform` hook on the four
   plugins to the object form.
2. Give each hook a filter that matches every id its handler can act on.
   The filter may match more ids than the handler accepts. It must never
   match fewer.
3. Keep every in-handler guard unchanged. The filter is an optimization.
   The guards remain the correctness boundary.

The filters, per plugin and hook:

| Plugin | Hook | Filter | Matches |
| --- | --- | --- | --- |
| `server-fn-client` | `transform` | include `/\.(ts\|tsx)$/`, exclude `/\/node_modules\//` and `/^\0/` | user `.ts`/`.tsx` files outside `node_modules` |
| `oj-assets` | `resolveId` | include `/\?(raw\|url\|inline)$/`, the asset-extension regex `ASSET_EXT`, `/\.css(\?\|$)/` | asset imports (`?raw`/`?url`/`?inline` suffixes, binary asset extensions, CSS) |
| `oj-assets` | `load` | `/^\0oj-(raw\|url\|inline\|css):/` | the virtual ids this plugin's own `resolveId` creates |
| `oj-vite-plugins` | `resolveId` | include `/\.svg\?react$/`, `/^virtual:/`, `/^\0/` | SVG-as-React imports and virtual ids that user Vite plugins may resolve |
| `oj-vite-plugins` | `load` | include `/^\0oj-/`, `/\.svg$/`, `/^(?!.*\/node_modules\/).*\.(jsx?\|mjs\|tsx?)(\?\|$)/` | oj virtual ids, SVG files, and user JS/TS files whose `load` a user Vite plugin may override |
| `oj-vite-plugins` | `transform` | `/\.mdx?$/` | Markdown/MDX files that user Vite plugins transform |
| `node-builtin-shims` | `resolveId` | include `/^node:/`, the builtin-name regex `BARE_BUILTINS` | Node.js builtin specifiers |
| `node-builtin-shims` | `load` | `/^\0oj-node-shim:/` | the shim virtual ids this plugin's own `resolveId` creates |

Two details need explanation:

- The `oj-vite-plugins` `load` filter uses a negative lookahead
  (`(?!.*\/node_modules\/)`) instead of an `exclude` entry. An `exclude`
  entry applies to the whole filter. The plugin's virtual ids (for example
  `\0oj-css:<absolute path>`) embed absolute paths, and those paths can
  contain `node_modules`. A global exclude would block them. The lookahead
  confines the `node_modules` exclusion to the user-file alternative only.
- `resolveId` filters accept RegExp only. Rolldown matches these patterns
  natively with the regress regex engine. We probe-verified that regress
  handles both `\0` literals and negative lookahead.

`crates/oj_server/src/assets/start/build.mjs` (the production build) imports
`oj-assets`, `oj-vite-plugins`, and `node-builtin-shims` from
`rolldown-assets.mjs`. The filters therefore apply to the production build
as well, under the same superset rule.

Files changed:

| Path | What changed |
| --- | --- |
| `crates/oj_server/src/assets/start/bundle-client.mjs` | `server-fn-client` `transform` converted to `{ filter, handler }`. Handler body unchanged. |
| `crates/oj_server/src/assets/start/rolldown-assets.mjs` | All `resolveId`/`load`/`transform` hooks on `oj-assets`, `oj-vite-plugins`, and `node-builtin-shims` converted to `{ filter, handler }`. Handler bodies unchanged. |

## Correctness

Every in-handler guard stays in place. A filter that matches too broadly
costs one JS call whose handler returns null. That is the behavior before
this change. A filter that matches too narrowly would change output. The
filters only skip calls whose handlers returned null, so output cannot
change. We verified this:

- We built the client bundle for the test app with and without the filters
  and hashed every emitted chunk. All 3,903 chunks were byte-identical.
- We booted the dev server with and without the filters. The sha of the
  served client entry was equal across boots.

Edge cases:

- Virtual ids that embed `node_modules` paths (`\0oj-css:`,
  `\0oj-svg-react:`): handled by the negative lookahead described above.
- `\0` bytes in filter patterns: `resolveId` filters accept RegExp only, and
  the native regress engine matches `\0` literals. Probe-verified.
- Hooks without an id to filter on (`buildStart` in `oj-vite-plugins`):
  unchanged, they keep the bare function form.

Not covered on purpose: no filter narrows beyond its handler's guards. We
did not tighten any guard, and we did not remove any guard.

## Performance

Measurement conditions: the 18,600-module test app above, macOS arm64,
release build of oj, warm `.oj-cache`, runs serialized (never concurrent).

- Bundle stage: 4.4–5.7 s before, 1.3–1.9 s after (approximately 3x).
- TTFC (wall-clock time from `oj dev` start until `GET /` returns HTTP 200
  with HTML): approximately 1 s faster. The bundle stage was not the
  critical path on this app, so most of the stage win does not reach TTFC.
- The same plugins run on every watch rebuild, so every rebuild gets the
  same per-module savings.

## How to test

2. Run `oj dev` in a TanStack Start app. Confirm `GET /` returns HTTP 200
   and the app renders.
3. Verify output identity: run `oj dev` once on `main` and once on this
   branch. After each boot, hash the bundled entry:
   `shasum .oj-cache/start/client-entry.js`. The hashes must be equal.
4. Verify the filters engage: time the "client entry bundling" step (oj
   prints it during boot) on both branches. On an app with many
   `node_modules` modules the step is measurably faster on this branch.

### DIFF
--- a/crates/oj_server/src/assets/css-preprocess.mjs
+++ b/crates/oj_server/src/assets/css-preprocess.mjs
@@ -2,6 +2,7 @@
 // Copyright (c) 2026 Raphael Amorim
 
 import { createRequire } from "node:module";
+import { fstatSync } from "node:fs";
 import { pathToFileURL } from "node:url";
 import path from "node:path";
 import readline from "node:readline";
@@ -28,6 +29,14 @@ async function stylus(base, css, from) {
 }
 
 const rl = readline.createInterface({ input: process.stdin });
+// stdin EOF means the parent is gone (SIGKILL skips its teardown) or the
+// one-shot build path closed it: exit once no reply is in flight.
+let inflight = 0;
+let stdinClosed = false;
+const maybeExit = () => { if (stdinClosed && inflight === 0) process.exit(0); };
+try {
+  if (fstatSync(0).isFIFO()) rl.once("close", () => { stdinClosed = true; maybeExit(); });
+} catch {}
 rl.on("line", async (line) => {
   let msg;
   try {
@@ -37,6 +46,7 @@ rl.on("line", async (line) => {
   }
   const { id, base, css, from } = msg;
   const ext = String(from || "").split("?")[0].split(".").pop().toLowerCase();
+  inflight += 1;
   try {
     let out = css;
     if (ext === "less") out = await less(base, css, from);
@@ -44,5 +54,8 @@ rl.on("line", async (line) => {
     process.stdout.write(JSON.stringify({ id, css: out }) + "\n");
   } catch (e) {
     process.stdout.write(JSON.stringify({ id, error: String((e && e.message) || e) }) + "\n");
+  } finally {
+    inflight -= 1;
+    maybeExit();
   }
 });
--- a/crates/oj_server/src/assets/plugin-host.mjs
+++ b/crates/oj_server/src/assets/plugin-host.mjs
@@ -2,7 +2,7 @@
 // Copyright (c) 2026 Raphael Amorim
 
 import http from "node:http";
-import { writeFileSync } from "node:fs";
+import { fstatSync, writeFileSync } from "node:fs";
 import { createRequire } from "node:module";
 import { pathToFileURL } from "node:url";
 import { dirname } from "node:path";
@@ -595,6 +595,15 @@ async function run(hook, args) {
   }
   return null;
 }
+
+// stdin EOF means oj died without tearing us down (SIGKILL skips
+// kill_on_drop): exit instead of idling as an orphan.
+try {
+  if (fstatSync(0).isFIFO()) {
+    process.stdin.once("end", () => process.exit(0));
+    process.stdin.once("close", () => process.exit(0));
+  }
+} catch {}
 
 const rl = readline.createInterface({ input: process.stdin });
 rl.on("line", async (line) => {
--- a/crates/oj_server/src/assets/ssr-runner.mjs
+++ b/crates/oj_server/src/assets/ssr-runner.mjs
@@ -3,7 +3,7 @@
 
 import vm from "node:vm";
 import readline from "node:readline";
-import { statSync } from "node:fs";
+import { fstatSync, statSync } from "node:fs";
 
 const [BASE, ENTRY] = process.argv.slice(2);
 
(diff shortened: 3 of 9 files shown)


raphamorim:
Landed on `main` as 837c940 (squashed, authored by @aeriksson-lovable). Verified each hook filter is a superset of its handler's positive path (every in-handler guard retained), so output is unchanged — the plugin-container routing tests (`asset-routing`, `rolldown-assets`, 15/15) pass. Thanks André!

PR #14, opened 2026-08-21
serve every chunk of the code-split client bundle

aeriksson-lovable:
Part 3 of 14 in a series. Depends on #13 — this PR's diff includes the commits of the PRs below it in the series; to see only this part: https://github.com/aeriksson-lovable/oj/compare/pr/02-rolldown-hook-filters...pr/03-serve-all-client-chunks. Series overview in #12.

---

## Problem

`oj dev` in start mode boots a TanStack Start application. During boot,
`crates/oj_server/src/assets/start/bundle-client.mjs` runs a rolldown
`build()` with `write: false` and persists exactly one output file:

```js
const chunk = result.output.find((o) => o.type === "chunk" && o.isEntry) ?? result.output[0];
writeFileSync(join(HERE, "client-entry.js"), chunk.code);
```

rolldown code-splits on dynamic imports. In a TanStack Start route tree,
every route component is a dynamic import, so a real application produces
many chunks. On our test app (18,600-module client graph; see Performance
for the full description), the build produces 3,903 output files, 80.1 MB
in total. The entry chunk is 14.4 KB of glue that statically imports 8
hoisted shared chunks (`rolldown-runtime-*.js`, `react-*.js`,
`client-*.js`, and 5 more). oj writes 1 of the 3,903 files and discards
the other 3,902.

The serve route in `crates/oj/src/start_dev.rs` knows four `/@oj-start/`
paths: `client-entry.js`, `live-reload.js`, `hmr`, and `fs/*`. There is no
handler and no on-disk file for any sibling chunk. The browser loads the
entry, requests the 8 static siblings, and receives 404 for every one.
Module resolution fails, so hydration never starts. Server-rendered HTML
and CSS still work, so the page renders but no client code runs: no
interactivity, no client-side navigation, no lazy routes.

The chunk hashes are content-derived, so every boot regenerates the same
names and the same 404s. A hard reload does not help.

We considered and rejected `output.codeSplitting: false`: it works (1
chunk, servable by the existing route), but the single dev chunk is
87.6 MB, every route loads eagerly, and the browser parses all of it on
first load.

## Change

The bundler persists every output file, and the server serves any file of
the current build under `/@oj-start/<name>`, resolving names only through
an index of that one build.

1. After the build, `bundle-client.mjs` writes every `result.output` file
   (chunks and assets) into `.oj-cache/start/client-chunks/`. It removes
   the previous build's directory first, so the directory always holds
   exactly one build.
2. It also writes `client-chunks.json`: the entry chunk's name plus every
   output file's name and size. This index defines the serve set.
3. The reported module count (`client-entry.modules`) now sums the modules
   of every chunk, not only the entry chunk (the entry holds ~11 modules
   after splitting; the real graph holds ~18,600).
4. On boot, the server reads the index once and pins it. A request for
   `/@oj-start/<name>` resolves through the pinned index: a hit serves the
   file with its MIME type; a name absent from the index is a deliberate
   404. Nothing outside the pinned build is reachable.
5. A watch rebuild writes a fresh directory and index, and swaps the
   pinned index atomically under a lock. In-flight requests finish against
   the old index; new requests see only the new build.
6. The `/@oj-start/fs` prefix match gains its trailing slash
   (`/@oj-start/fs/`). Without it, the filesystem-asset route shadows any
   chunk whose name starts with `fs` (measured: `fsharp-D7o3fj_O.js`
   returned 404 through the wrong handler).

Files changed:

| Path | What changed and why |
| --- | --- |
| `crates/oj_server/src/assets/start/bundle-client.mjs` | Writes all output files and the index; module count sums all chunks. |
| `crates/oj/src/start_dev.rs` | Generic `/@oj-start/<name>` route resolving through the pinned index; `fs/` prefix fix; the pinned index lives in `StartState` and swaps on rebuild. |

## Correctness

**Names resolve only through the pinned index.** The route does a map
lookup on the exact name; path traversal (`../`) cannot match a key, and a
name from a different build is a 404, never a stale file. The index and
the files it names come from the same build, so the serve set can never
mix two builds.

**Rebuild swap is atomic.** The pinned index is swapped whole under a
lock after the rebuild finishes. The reload event that makes the browser
refetch is sent after the swap.

**Unknown extensions** serve as `application/octet-stream`; `.js`
chunks serve as `text/javascript` (the existing `asset_mime` table).

**Intentionally not covered.** Responses keep `Cache-Control: no-cache`,
as the entry did before; conditional requests (ETag) are a possible
follow-up, not part of this fix.

## Performance

concurrently.

- Build output on this app: 3,903 files, 80.1 MB. Before this change, 1
  file was written and 3,902 were discarded; the browser received 404 for
  every chunk the 14.4 KB entry imports, and hydration never ran.
- After this change, a fetch battery requested `GET /` plus every one of
  the 3,903 names in the index: all returned 200 with byte-exact content,
  85.6 MB in 0.6 s over 16 connections (server on localhost).
- The transitive static/dynamic import closure reachable from the served
  entry resolves entirely inside the index (3,917 specifier references,
  0 unresolved).
- Writing 3,903 files instead of 1 does not change TTFC: cold 8 s,
  warm 4 s / 3 s — the same ladder as the previous binary on this app
  (cold 8 s, warm 5 s / 3 s), because the write happens inside the
  existing bundle step off the request path.
- Initial page load now fetches the entry plus 8 static siblings
  (~15 MB) instead of failing; lazy routes stay lazy.

## How to test

2. Run `oj dev` in a TanStack Start app whose routes use dynamic imports
   (any generated route tree qualifies).
3. Open the app in a browser. Before this change, the network tab shows
   404 for every `/@oj-start/<name>-<hash>.js` sibling and the page never
   hydrates. After it, all chunk requests return 200 and the app is
   interactive.
4. Request a name that is not part of the build, e.g.
   `curl -i http://localhost:<port>/@oj-start/nope.js` — expect 404.
5. Edit a route file; after the rebuild the browser reloads and the new
   chunk set serves.

### DIFF
--- a/crates/oj/src/start_dev.rs
+++ b/crates/oj/src/start_dev.rs
@@ -25,7 +25,10 @@ struct StartState {
     proxy_prefixes: Vec<String>,
     plugin_mw_port: Option<u16>,
     runner: Arc<tokio::sync::Mutex<Runner>>,
-    client_bundle: PathBuf,
+    // The chunk set of the client build this server serves: swapped whole on
+    // watch rebuilds, so every /@oj-start/<chunk> resolves against exactly
+    // one build and a name outside it is a deliberate 404.
+    bundle: std::sync::RwLock<Arc<PinnedBundle>>,
     live_reload: PathBuf,
     workspace_root: PathBuf,
     reload_tx: broadcast::Sender<()>,
@@ -60,7 +63,12 @@ pub async fn start_dev(
     route_tree.await??;
     let bundle = {
         let (root, cache) = (root.clone(), cache.clone());
-        tokio::task::spawn_blocking(move || bundle_client_entry(&root, &cache))
+        tokio::task::spawn_blocking(move || -> anyhow::Result<PinnedBundle> {
+            bundle_client_entry(&root, &cache)?;
+            PinnedBundle::from_build_dir(&cache).ok_or_else(|| {
+                anyhow::anyhow!("client chunk index missing after successful bundle")
+            })
+        })
     };
     let built_fut = oj_server::DevServer {
         root: root.clone(),
@@ -71,7 +79,7 @@ pub async fn start_dev(
     }
     .build_app();
     let (bundle_res, built_res) = tokio::join!(bundle, built_fut);
-    bundle_res??;
+    let pinned = bundle_res??;
     resolver.await??;
     let built = built_res?;
     let runner = spawn_start_runner(&root, &cache).await?;
@@ -88,7 +96,7 @@ pub async fn start_dev(
         proxy_prefixes: built.proxy_prefixes.clone(),
         plugin_mw_port: built.plugin_mw_port,
         runner: Arc::new(tokio::sync::Mutex::new(runner)),
-        client_bundle: cache.join("client-entry.js"),
+        bundle: std::sync::RwLock::new(Arc::new(pinned)),
         live_reload: cache.join("live-reload.js"),
         workspace_root: workspace_root(&root),
         reload_tx: reload_tx.clone(),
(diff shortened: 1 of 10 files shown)


raphamorim:
Landed on `main` as 6cd3ca8 (squashed, authored by @aeriksson-lovable). The `PinnedBundle` swap-whole design keeps the serve set from ever mixing two builds, the `client-entry.js` fallback preserves the entry URL, and the `/@oj-start/fs/` trailing slash stops `fs`-prefixed chunks being shadowed. Verified `e2e/start.mjs` (dev + prod client serving) and `ssr-dev` pass. Thanks André!

PR #15, opened 2026-08-21
cache start-mode codegen, enable the V8 compile cache on node children

aeriksson-lovable:
Part 6 of 16 in a series. Depends on #27 — this PR's diff includes the commits of the PRs below it in the series; to see only this part: https://github.com/aeriksson-lovable/oj/compare/pr/03c-nested-require-resolution...pr/04-codegen-cache. Series overview in #12.

---

## Problem

`oj dev` in TanStack Start mode runs two code generators before it can bundle.
Both run on every boot.

1. Route-tree generation. `start_dev` in `crates/oj/src/start_dev.rs` spawns a
   node process that runs the embedded script
   `crates/oj_server/src/assets/start/generate.mjs`. The script runs the
   TanStack router-generator and writes `src/routeTree.gen.ts` in the app.
2. Server-function resolver generation. `start_dev` spawns a second node
   process that runs `crates/oj_server/src/assets/start/gen-resolver.mjs`. The
   script scans `src/**/*.ts(x)` for `createServerFn` calls and writes a
   resolver module into `.oj-cache/start/`.

The client bundle waits on the route tree, so step 1 sits on the critical path
to the first response. The two generator runs cost about 0.4–0.5 s per boot
with a hot OS file cache on our test app (described under Performance). The
file watcher also reruns the resolver on every batch that touches a route file
or a server-function file.

Separately, every node child oj spawns compiles its JavaScript from scratch on
every boot. Node ships a built-in V8 bytecode cache (`NODE_COMPILE_CACHE`),
but oj does not enable it.

## Change

Two independent parts.

### Part 1: a content-addressed cache for the two generator outputs

New module `crates/oj_cache/src/start_codegen.rs` adds `StartCodegenStore`.
The store keeps generator outputs under
`.oj-cache/start-codegen/<kind>/<key>/`, where `kind` is `route-tree` or
`server-fn` and `key` is a blake3 hash of the generator's inputs.

The restore path, on every boot:

1. The caller lists the candidate input files. For the route tree that is
   every `.ts`/`.tsx` file under `src/routes/`. For the resolver that is every
   `.ts`/`.tsx` file under `src/`.
2. The store sorts and dedups the paths, so the key does not depend on
   directory traversal order.
3. The store obtains a blake3 digest for each input. To avoid reading every
   file on every boot, it keeps a per-kind stat memo (`memo.json`): a map from
   relative path to (size, mtime in ns, digest, marked flag). A file whose
   current size and mtime match the memo reuses the stored digest. Any other
   file is read and re-hashed. The memo is advisory only: content stays
   authoritative, so a stale memo can cause extra reads, never a wrong digest.
   The digest pass runs across threads (one chunk per core).
4. The memo applies git's racy-index rule. When the store writes the memo, it
   stamps the write time and leaves out every file whose mtime is within 2 s
   of that stamp. Reason: a write that lands in the same timestamp tick as the
   stat can change content without changing (size, mtime). An omitted file is
   re-hashed on each boot until a later memo write sees its mtime settled.
5. The store computes the key: blake3 over a salt, then, for each keyed file
   in sorted order, the relative path and the content digest. The salt binds
   the entry to the oj version, a format number, the kind, the generator
   script bytes, and the contents of `tsr.config.json`, `package.json`, and
   the lockfiles (`package-lock.json`, `yarn.lock`, `pnpm-lock.yaml`,
   `bun.lockb`) when present. Upgrading oj, the script, the router config, or
   a dependency therefore changes the key.
6. Which files are "keyed" depends on the kind. The route tree keys every
   input. The resolver has a marker, the string `createServerFn`: only files
   that contain the marker feed the key. Editing a file without the marker
   still hits; a file that gains or loses the marker changes the key, because
   the marked flag is recomputed whenever a file is re-hashed and adding or
   removing a keyed file changes the hashed sequence.
7. If `.oj-cache/start-codegen/<kind>/<key>/` exists, the store copies each
   entry file to its destination and the boot skips the generator. If the
   destination already holds byte-identical content, the store skips the
   write. If any entry file cannot be read, the store deletes the entry and
   reports a miss.

Why the byte-identical check matters: an unchanged destination keeps its
mtime. `src/routeTree.gen.ts` lives under `src/`, so it is itself an input to
the resolver's stat memo, and other stat-based caches and file watchers see it
too. A rewrite with identical bytes would bump the mtime every boot and force
those consumers to re-read or re-run for no reason. Skipping the write keeps
every downstream stat check settled.

The miss path:

1. The generator runs exactly as before this change.
2. The store then hashes the inputs again, post-run, and persists the outputs
   under the resulting key.

Why post-run hashing is required for the route tree: the TanStack
router-generator normalizes route files in place while it runs (for example,
it rewrites `createFileRoute` ids). A key over the pre-run contents would
describe a file state that never exists again — the very next boot already
sees the normalized files, computes a different key, and misses forever. The
post-run contents are the generator's fixed point: running the generator on
already-normalized files leaves them unchanged, so the next boot reproduces
the same key and hits.

Housekeeping and failure behavior:

- Entries are written to a temporary directory and renamed into place. A lost
  rename race means another process persisted the same key; that is fine.
- Each restore or persist touches a `last-used` file inside the entry. A prune
  step keeps the 8 most recently used entries per kind and deletes the rest,
  so switching between branches keeps both branches' entries warm.
- Every failure degrades to a miss: unreadable input, missing entry, corrupt
  entry (deleted on sight), corrupt memo (deleted on sight), failed persist.
  No cache failure can break the boot or serve stale output.

`crates/oj/src/start_dev.rs` wraps both generators with the store, in
`generate_route_tree` and the new `generate_server_fn_resolver`. The dev boot,
the watcher, and `oj build` (start mode) all go through these wrappers. Hits
and misses print one line each, with the short key, the keyed-file count, the
re-hash count, and the elapsed time.

Known limitation: the plugin host's TanStack router plugin rewrites
`src/routeTree.gen.ts` during the bundle, which bumps its mtime every boot.
The resolver therefore re-hashes that one file on every boot. The content is
unchanged, so the key is unchanged and the cache still hits; the cost is one
file read.

### Part 2: enable node's V8 compile cache on node children

New helpers in `crates/oj_server/src/lib.rs`:

- `node_compile_cache(root)` returns the user's own `NODE_COMPILE_CACHE` value
  when set, else `<root>/.oj-cache/v8`. A user-set value always wins.
- `node_compile_cache_opt_in(root)` returns a path only when the user sets
  `OJ_V8_COMPILE_CACHE` to a non-empty value other than `0`.

Every short-lived or restartable node child now gets
`NODE_COMPILE_CACHE=<root>/.oj-cache/v8`: the two codegen scripts, the client
bundler, the production build and prerender, the css/svelte/preprocess
sidecars, the optimizer, the vite-config extractor, and the plugin host.

Three services are excluded by default and use the opt-in helper instead: the
SSR runner in `crates/oj/src/ssr_dev.rs`, the start-mode runner, and the
css host. On the test app, a V8 cache in those
processes measured as a net time cost, not a saving, so they stay uncached
unless the user sets `OJ_V8_COMPILE_CACHE`.

### Files changed

| Path | What changed |
| --- | --- |
| `crates/oj_cache/src/start_codegen.rs` | New. `StartCodegenStore`: keying, stat memo, racy-tick rule, atomic restore/persist, touch-LRU prune, unit tests. |
| `crates/oj_cache/src/lib.rs` | Exports the new module. |
| `crates/oj/src/start_dev.rs` | Wraps route-tree and resolver generation with the store. Adds `list_src_ts_files`. Sets `NODE_COMPILE_CACHE` in `run_node` and the production build; opt-in for the start runner and css host. |
| `crates/oj/src/ssr_dev.rs` | Opt-in compile cache for the SSR runner. |
| `crates/oj/src/build.rs` | Compile cache on the css, svelte, preprocess, and prerender node spawns. |
| `crates/oj_server/src/lib.rs` | Adds `node_compile_cache` and `node_compile_cache_opt_in`. |
| `crates/oj_server/src/optimize.rs` | Compile cache on the optimizer spawn. |
| `crates/oj_server/src/plugins.rs` | Compile cache on the config extractor and plugin host spawns. |
| `crates/oj_server/src/sidecar.rs` | Compile cache on sidecar spawns. |
| `crates/oj/Cargo.toml`, `Cargo.lock` | `oj` now depends on `oj_cache`. |
| `e2e/start.mjs` | Retries `fs.rmSync` on the temp app. A killed node child can still flush its compile cache into `.oj-cache/v8` after the kill, which races the recursive removal (ENOTEMPTY). |

## Correctness

What invalidates the cache:

- Any content change to a keyed input file (route file for the route tree, a
  `createServerFn` file for the resolver). The key is content-addressed, so a
  touch without a content change re-hashes to the same key and still hits.
- Adding or removing a keyed input file.
- A file gaining or losing the `createServerFn` marker.
- A change to the generator script, `tsr.config.json`, `package.json`, any
  lockfile, the oj version, or the store format number (all in the salt).

Why the stat memo cannot cause a stale hit: the memo only short-circuits the
digest of a file whose size and mtime both match. The one hole in that check
is an edit inside the same timestamp tick as the recorded stat. The racy-tick
rule closes it: files with an mtime within 2 s of the memo's write time are
never recorded, so they are always re-hashed. This is the same rule git uses
for its index. A unit test constructs the attack directly (same size, mtime
reset to the recorded value, different content) and asserts a miss.

Failure handling, per case:

- Input file deleted or unreadable → miss, generator runs.
- Entry directory missing for the computed key → miss, generator runs.
- Entry file missing or unreadable → entry deleted, miss, generator runs.
- `memo.json` corrupt → memo deleted, every input re-hashed, correct key.
- Persist fails at any step → temp directory removed, store unchanged.
- Two processes persist the same key → the rename loser accepts the winner's
  entry; the content is the same by construction (same key, deterministic
  generators over the same inputs).

Intentionally not covered:

- The resolver's input set (every `.ts`/`.tsx` under `src/`) over-approximates
  what the resolver reads. Over-approximation can only add key sensitivity; it
  can produce an unnecessary miss, never a stale hit.
- Generator inputs outside `src/` and outside the salted config files (for
  example, an env variable that changes generator behavior) are not keyed. The
  current scripts read no such inputs.
- The cache has no dedicated kill switch. `rm -rf .oj-cache/start-codegen`
  resets it, and every failure already degrades to the pre-change behavior.
- `.oj-cache/v8` is managed by node itself (it validates entries against the
  source and its own version), so oj adds no invalidation logic for it.

## Performance

(M-series), release build of oj. TTFC means the wall-clock time from `oj dev`
start until `GET /` returns HTTP 200 with HTML. Warm means the `.oj-cache`
directory is populated from a previous run. Cold means the cache directory is
empty. All measurements ran serially, never concurrently.

- Before: the two generator runs cost about 0.4–0.5 s per boot (hot OS file
  cache), and the route tree gates the client bundle.
- After, warm: the route-tree phase drops from 0.36–0.48 s to 0.016–0.06 s.
  The store verifies the 492 route files of this app in 1–2 ms via the stat
  memo.
- The V8 compile cache contributes a further ~0.9–1.3 s per boot on this app,
  measured as a separate contribution on the node children listed above.
- Negative result: enabling the V8 compile cache in the runners and the
  css host measured as a net cost per boot on this app. Those services are
  therefore excluded by default and gated behind `OJ_V8_COMPILE_CACHE`.
- Cold boots pay one extra hash pass over the input files and one entry write.
  Measured cold TTFC on this app, against a stock 0.0.45 control run in the
  same session: 13 s without this change, 15 s with it
  (one cold run per arm). Warm TTFC stayed at 12 s in both arms (two warm
  runs per arm); the phase-level savings above sit below the 1 s resolution
  of the TTFC ladder.

## How to test

Unit tests:

```
cargo test -p oj_cache start_codegen
cargo test -p oj start_dev
```

End to end (builds oj, boots the fixture app twice per scenario):

```
node e2e/start.mjs
```

Manual, in any TanStack Start app:

1. Build oj and run `oj dev`. The log prints
   `route tree cache miss (no cached output for key …)` and the generators
   run.
2. Stop the server. Run `oj dev` again. The log prints
   `route tree restored from cache (key …, N route file(s) verified in Xms)`
   and `server-fn resolver restored from cache (…)`.
3. Edit a route file. The next boot misses, regenerates, and persists a new
   entry. Revert the edit. The next boot hits the older entry again.
4. Edit a `src/` file that does not contain `createServerFn`. The resolver
   still hits. Add a `createServerFn` call to it. The resolver misses.
5. Check `ls .oj-cache/start-codegen/route-tree/` after many edits: at most 8
   entries plus `memo.json` remain.
6. Check `ls .oj-cache/v8/`: node writes `.cache` files there after any boot.
7. Confirm `git status` in the app stays clean across warm boots: the restore
   does not rewrite an unchanged `src/routeTree.gen.ts`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1794,6 +1794,7 @@ dependencies = [
  "axum",
  "clap",
  "notify",
+ "oj_cache",
  "oj_compiler",
  "oj_config",
  "oj_css",
--- a/crates/oj/Cargo.toml
+++ b/crates/oj/Cargo.toml
@@ -10,6 +10,7 @@ name = "oj"
 path = "src/main.rs"
 
 [dependencies]
+oj_cache.workspace = true
 oj_compiler.workspace = true
 oj_env.workspace = true
 oj_config.workspace = true
--- a/crates/oj/src/build.rs
+++ b/crates/oj/src/build.rs
@@ -583,6 +583,7 @@ fn expand_css_via_sidecar(root: &Path, css_file: &Path) -> anyhow::Result<String
             css_file.to_str().unwrap(),
             root.to_str().unwrap(),
         ])
+        .env("NODE_COMPILE_CACHE", oj_server::node_compile_cache(root))
         .current_dir(root)
         .output()
         .context("node not found for tailwind/postcss build")?;
@@ -614,6 +615,7 @@ fn svelte_via_sidecar(root: &Path, file: &Path) -> anyhow::Result<String> {
     .to_string();
     let mut child = std::process::Command::new("node")
         .arg(&script)
+        .env("NODE_COMPILE_CACHE", oj_server::node_compile_cache(root))
         .current_dir(root)
         .stdin(std::process::Stdio::piped())
         .stdout(std::process::Stdio::piped())
@@ -658,6 +660,7 @@ fn preprocess_via_sidecar(root: &Path, css_file: &Path) -> anyhow::Result<String
     .to_string();
     let mut child = std::process::Command::new("node")
         .arg(&script)
+        .env("NODE_COMPILE_CACHE", oj_server::node_compile_cache(root))
         .current_dir(root)
         .stdin(std::process::Stdio::piped())
         .stdout(std::process::Stdio::piped())
@@ -1974,6 +1977,7 @@ pub(crate) async fn build_ssr_app(
         let out = std::process::Command::new("node")
             .arg(&script_path)
             .arg(serde_json::to_string(&paths)?)
+            .env("NODE_COMPILE_CACHE", oj_server::node_compile_cache(root))
             .current_dir(out_dir)
             .output()
             .context("node not found for prerender")?;
--- a/crates/oj/src/ssr_dev.rs
+++ b/crates/oj/src/ssr_dev.rs
@@ -86,17 +86,21 @@ async fn spawn_runner(root: &Path, base: &str, entry_abs: &Path) -> anyhow::Resu
     let script = dir.join("runner.mjs");
     std::fs::write(&script, oj_server::SSR_RUNNER_JS)?;
 
-    let mut child = tokio::process::Command::new("node")
-        .args([
-            "--experimental-vm-modules",
-            &script.to_string_lossy(),
-            base,
-            &entry_abs.to_string_lossy(),
-        ])
-        .current_dir(root)
-        .stdin(Stdio::piped())
-        .stdout(Stdio::piped())
-        .stderr(Stdio::inherit())
+    let mut cmd = tokio::process::Command::new("node");
+    cmd.args([
+        "--experimental-vm-modules",
+        &script.to_string_lossy(),
+        base,
+        &entry_abs.to_string_lossy(),
+    ])
+    .current_dir(root)
+    .stdin(Stdio::piped())
+    .stdout(Stdio::piped())
+    .stderr(Stdio::inherit());
+    if let Some(v8) = oj_server::node_compile_cache_opt_in(root) {
+        cmd.env("NODE_COMPILE_CACHE", v8);
+    }
+    let mut child = cmd
         .spawn()
         .map_err(|e| anyhow::anyhow!("could not spawn SSR runner (node): {e}"))?;
 
--- a/crates/oj/src/start_dev.rs
+++ b/crates/oj/src/start_dev.rs
@@ -25,7 +25,10 @@ struct StartState {
     proxy_prefixes: Vec<String>,
     plugin_mw_port: Option<u16>,
     runner: Arc<tokio::sync::Mutex<Runner>>,
-    client_bundle: PathBuf,
+    // The chunk set of the client build this server serves: swapped whole on
+    // watch rebuilds, so every /@oj-start/<chunk> resolves against exactly
+    // one build and a name outside it is a deliberate 404.
+    bundle: std::sync::RwLock<Arc<PinnedBundle>>,
     live_reload: PathBuf,
     workspace_root: PathBuf,
     reload_tx: broadcast::Sender<()>,
@@ -60,7 +63,12 @@ pub async fn start_dev(
     route_tree.await??;
     let bundle = {
         let (root, cache) = (root.clone(), cache.clone());
-        tokio::task::spawn_blocking(move || bundle_client_entry(&root, &cache))
+        tokio::task::spawn_blocking(move || -> anyhow::Result<PinnedBundle> {
+            bundle_client_entry(&root, &cache)?;
+            PinnedBundle::from_build_dir(&cache).ok_or_else(|| {
+                anyhow::anyhow!("client chunk index missing after successful bundle")
+            })
+        })
     };
     let built_fut = oj_server::DevServer {
         root: root.clone(),
@@ -71,7 +79,7 @@ pub async fn start_dev(
     }
     .build_app();
     let (bundle_res, built_res) = tokio::join!(bundle, built_fut);
-    bundle_res??;
+    let pinned = bundle_res??;
     resolver.await??;
     let built = built_res?;
     let runner = spawn_start_runner(&root, &cache).await?;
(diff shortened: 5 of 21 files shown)


raphamorim:
Landed on `main` as 565f510 (squashed, authored by @aeriksson-lovable). Reviewed the cache design: key = blake3(salt + each marked input's rel-path + content digest); the stat memo trusts (size, mtime) but content stays authoritative on the miss path, and the git racy-window rule (2s) prevents memoizing a file that could change within stat granularity, so it can cause extra reads but never a false hit. `copy_atomic` skips byte-identical writes to keep downstream mtimes settled; persist is tmp+rename atomic with touch-LRU prune to 8. `oj_cache` (incl. the 12 start_codegen tests) and `e2e/start.mjs` (dev + prod) pass. Thanks André!

PR #16, opened 2026-08-21
add a persistent cache for the start client bundle

aeriksson-lovable:
Part 7 of 16 in a series. Depends on #15 — this PR's diff includes the commits of the PRs below it in the series; to see only this part: https://github.com/aeriksson-lovable/oj/compare/pr/04-codegen-cache...pr/05-client-bundle-cache. Series overview in #12.

---

## Problem

`oj dev` in start mode boots a TanStack Start application. During boot,
`start_dev` in `crates/oj/src/start_dev.rs` spawns a Node.js process that
runs `crates/oj_server/src/assets/start/bundle-client.mjs`. This script
runs a full rolldown build of the client entry and writes the output into
`.oj-cache/start/`: the code-split chunk set under `client-chunks/`
(3,903 files, 80.1 MB on our test app), an index of those files, a module
count, and the dev manifest.

The build runs on every boot. It runs even when no input file changed
since the previous boot. On our test app (18,600-module client graph,
macOS arm64, release build of oj), this build costs approximately 7 s
wall clock and 17 s CPU per boot.

A cache for this output has two hazards a small cache does not have.
First, the serve set is thousands of files: a restore that copies them
pays for the full 80 MB on every boot. Second, a serve set can tear: if
the entry file comes from build A and a sibling chunk from build B, the
browser loads mismatched code. The design below makes both structurally
impossible instead of unlikely.

## Change

Two terms used below. The closure is the list of files whose content
shaped one build. A generation is one cached build.

The store (`crates/oj_cache/src/start_bundle.rs`, `StartBundleStore`)
owns `.oj-cache/start-bundle/`:

```
start-bundle/
  blobs/<blake3-of-content>     one file per unique chunk content
  <key>/                        one directory per generation
    manifest.json               {format, entry, cssUrls, files: {name -> {hash, size}}}
    closure.json  memo.json     key inputs (as before)
    client-entry.modules  manifest.ts   small artifacts copied on restore
  current                       pointer to the pinned generation (rename-swapped)
```

Steps of the algorithm:

1. The bundler script records the traversed module graph (every id the
   transform pipeline visits, every chunk's rendered modules, the vite
   config's dependency list, `addWatchFile` calls) and writes it to
   `closure.json`. On our test app the closure holds 18,646 files.
2. The cache key is a blake3 hash over a salt followed by each closure
   file's (path, blake3(content)) pair. The salt contains the oj version,
   the store format number (`START_BUNDLE_FORMAT`), and an epoch hashing
   the lockfiles, `package.json`, the vite and oj config files, the
   `.env*` files, and the env vars the bundler inlines (`VITE_*`,
   `TSS_SERVER_FN_BASE`).
3. `persist` runs after every successful build. It hashes every chunk
   file and copies each into `blobs/` under its own hash — a blob that
   already exists is not written again, so generations share identical
   chunks. It then writes the generation manifest, builds the generation
   directory in a temp dir renamed into place, and swaps the `current`
   pointer by rename. The manifest is the single file that defines which
   names exist and which bytes they resolve to.
4. On boot, `restore` reads `current`, verifies the closure via a
   per-file (size, mtime, digest) stat memo (git's racy-index rule, 2 s
   freshness slack; a stat mismatch re-hashes, so the memo can cause
   extra reads, never a false hit), recomputes the key, and pins the
   matching generation: it verifies every manifest member against the
   blob store, copies only the two small artifacts into
   `.oj-cache/start/`, derives `css-urls.json` from the manifest, and
   returns the pinned name→blob map to the server. No chunk bytes are
   copied; the serve route reads blobs directly through the pinned map.
5. Member verification has two depths. Default: existence + size (a stat
   pass). `OJ_CACHE_VERIFY=full`: re-hash every blob against its manifest
   hash. Any failure deletes the failing blob and the generation and
   degrades to a rebuild — a verification failure can cost a rebuild,
   never wrong bytes.
6. The serve path resolves `/@oj-start/<name>` only through the pinned
   manifest. Reads go through `verified_read` (size always checked;
   content re-hashed under `OJ_CACHE_VERIFY=full`); a read that fails
   verification returns an error, not the bytes. A name absent from the
   pinned generation is a deliberate 404, so a serve set can never mix
   two builds.
7. Watch rebuilds persist a new generation and swap the server's pinned
   map whole. If `persist` fails, the server pins the fresh build
   directory directly (`PinnedBundle::from_build_dir`) — the store is an
   optimization, never a correctness dependency.
8. After boot, off the critical path, `prune` evicts generations
   oldest-first by last-use down to a 1 GB budget, counting each blob
   once (generations share blobs), never evicting the `current`
   generation, and then sweeps blobs no surviving generation references.
9. Shared primitives live in `crates/oj_cache/src/integrity.rs`:
   `atomic_write` (same-directory temp + rename), `verify_file`,
   `verified_read`, and the `OJ_CACHE_VERIFY` mode switch. Other stores
   in the crate can adopt the same contract.

Files changed:

| Path | What changed and why |
| --- | --- |
| `crates/oj_cache/src/integrity.rs` | New. Shared primitives: atomic writes, verified reads, verify modes. |
| `crates/oj_cache/src/start_bundle.rs` | New store: blob ingestion, generation manifest, pin/restore/persist, dedupe-aware LRU prune, memo; 20 unit tests. |
| `crates/oj_cache/src/lib.rs` | Exports the new modules. |
| `crates/oj/src/start_dev.rs` | Boot restores-or-builds and pins; the chunk route reads through the pinned map with `verified_read`; watch rebuilds persist and swap; `prune` after boot. |
| `crates/oj_server/src/assets/start/bundle-client.mjs` | Writes `closure.json` (recorder plugin) next to the chunk output of #14. |

## Correctness

**Why the closure is the traversed graph, not the rendered chunks.**
Traversal shapes the output (the manifest's `cssUrls` come from css
imports found anywhere in the graph), so the recorder plugin captures
every visited id, 18,646 files on our test app. Derived files
(`manifest.ts`, `routeTree.gen.*`, files under `.oj-cache/start/`) are
excluded: keying on outputs makes the key change without a source change
and turns the cache into an eternal miss.

**A generation cannot tear.** The manifest, the blobs it names, and the
small artifacts are committed by two renames (generation directory, then
`current`). The serve path resolves names only through one pinned
manifest, so it can never observe half of build A and half of build B.
A blob is content-addressed: its name is the hash of its bytes, so a
stale or foreign blob cannot satisfy a manifest entry under full verify,
and a wrong-size one fails even the default check.

**Every failure degrades to a miss.** A missing pointer, an unreadable
closure or manifest, a wrong format number, an absent generation, a
missing/wrong-size/wrong-hash blob: each returns a typed miss reason,
removes the bad entry, and the boot builds as before. A failed `persist`
leaves the store as it was and the server serves the fresh build
directory. No cache-IO error can fail the boot.

**Format versioning.** `START_BUNDLE_FORMAT` is in the key salt and in
the manifest. Entries written by an older format are unreachable by key
and rejected by the manifest check — invalid by version, not by
accident.

**Reverts hit older generations.** The `current` pointer names the
closure to verify, but the key comes from current file contents. After
an edit and a revert, the recomputed key lands on the older generation
and the pointer swings back to it.

**Intentionally not covered.** Default verification trusts a blob whose
size matches (same-size corruption needs `OJ_CACHE_VERIFY=full`), and
the key does not hash ambient module resolution (a new file that changes
how an unchanged specifier resolves is caught by the next edit or watch
rebuild, not at boot).

**Tests.** 20 unit tests in `start_bundle.rs` (roundtrip, pin/swap on
edit and revert, deliberate 404 for absent names, blob dedupe, truncated
blob caught in default mode, same-size bit flip caught in full mode and
healed by rebuild, format invalidation, legacy-entry miss, dedupe-aware
prune + blob sweep, and the memo suite) plus 5 in `integrity.rs`.

## Performance

concurrently.

- Warm restore of the full 3,903-chunk generation: 18,646 closure files
  plus 3,903 manifest members verified in 30 ms (default mode). No chunk
  bytes are copied on restore. The pre-manifest store verified the same
  closure in 26 ms, so serving the whole chunk set through the store
  costs ~4 ms of stat pass at boot.
- TTFC ladder (cold + warm ×2, this app): 8 s / 4 s / 3 s — identical to
  the same binary without the chunk store (8 s / 5 s / 3 s).
- Blob store size for one generation: 3,902 blobs (two chunks share
  content), 91 MB on disk.
- Cold boot (fresh build + persist of 3,903 chunks): 8 s TTFC, the same
  as before the store existed; hashing and copying 80 MB into the blob
  store happens inside the bundle step.
- Corruption drills on this app: one flipped byte in a blob is detected
  under `OJ_CACHE_VERIFY=full` at boot, the generation rebuilds, and the
  served bytes match the new manifest; a truncated blob is detected in
  default mode the same way.

## How to test

2. In a TanStack Start app, run `oj dev` once and stop it. The log prints
   `oj start: client bundle cached (key XXXXXXXX…, N chunks)`.
3. Run `oj dev` again. The log prints `oj start: client bundle restored
   from cache (key …, N chunks, M files verified in K ms)` and the
   bundle phase disappears from the boot.
4. Edit a route file and boot again: a miss, a build, a new generation.
   Revert and boot: the older generation restores.
5. Flip one byte in a file under `.oj-cache/start-bundle/blobs/` and boot
   with `OJ_CACHE_VERIFY=full`: the log prints the chunk-corrupt miss and
   the boot rebuilds.
6. Run the unit tests: `cargo test -p oj_cache` and `cargo test -p oj`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1794,6 +1794,7 @@ dependencies = [
  "axum",
  "clap",
  "notify",
+ "oj_cache",
  "oj_compiler",
  "oj_config",
  "oj_css",
--- a/crates/oj/Cargo.toml
+++ b/crates/oj/Cargo.toml
@@ -10,6 +10,7 @@ name = "oj"
 path = "src/main.rs"
 
 [dependencies]
+oj_cache.workspace = true
 oj_compiler.workspace = true
 oj_env.workspace = true
 oj_config.workspace = true
--- a/crates/oj/src/build.rs
+++ b/crates/oj/src/build.rs
@@ -583,6 +583,7 @@ fn expand_css_via_sidecar(root: &Path, css_file: &Path) -> anyhow::Result<String
             css_file.to_str().unwrap(),
             root.to_str().unwrap(),
         ])
+        .env("NODE_COMPILE_CACHE", oj_server::node_compile_cache(root))
         .current_dir(root)
         .output()
         .context("node not found for tailwind/postcss build")?;
@@ -614,6 +615,7 @@ fn svelte_via_sidecar(root: &Path, file: &Path) -> anyhow::Result<String> {
     .to_string();
     let mut child = std::process::Command::new("node")
         .arg(&script)
+        .env("NODE_COMPILE_CACHE", oj_server::node_compile_cache(root))
         .current_dir(root)
         .stdin(std::process::Stdio::piped())
         .stdout(std::process::Stdio::piped())
@@ -658,6 +660,7 @@ fn preprocess_via_sidecar(root: &Path, css_file: &Path) -> anyhow::Result<String
     .to_string();
     let mut child = std::process::Command::new("node")
         .arg(&script)
+        .env("NODE_COMPILE_CACHE", oj_server::node_compile_cache(root))
         .current_dir(root)
         .stdin(std::process::Stdio::piped())
         .stdout(std::process::Stdio::piped())
@@ -1974,6 +1977,7 @@ pub(crate) async fn build_ssr_app(
         let out = std::process::Command::new("node")
             .arg(&script_path)
             .arg(serde_json::to_string(&paths)?)
+            .env("NODE_COMPILE_CACHE", oj_server::node_compile_cache(root))
             .current_dir(out_dir)
             .output()
             .context("node not found for prerender")?;
--- a/crates/oj/src/ssr_dev.rs
+++ b/crates/oj/src/ssr_dev.rs
@@ -86,17 +86,21 @@ async fn spawn_runner(root: &Path, base: &str, entry_abs: &Path) -> anyhow::Resu
     let script = dir.join("runner.mjs");
     std::fs::write(&script, oj_server::SSR_RUNNER_JS)?;
 
-    let mut child = tokio::process::Command::new("node")
-        .args([
-            "--experimental-vm-modules",
-            &script.to_string_lossy(),
-            base,
-            &entry_abs.to_string_lossy(),
-        ])
-        .current_dir(root)
-        .stdin(Stdio::piped())
-        .stdout(Stdio::piped())
-        .stderr(Stdio::inherit())
+    let mut cmd = tokio::process::Command::new("node");
+    cmd.args([
+        "--experimental-vm-modules",
+        &script.to_string_lossy(),
+        base,
+        &entry_abs.to_string_lossy(),
+    ])
+    .current_dir(root)
+    .stdin(Stdio::piped())
+    .stdout(Stdio::piped())
+    .stderr(Stdio::inherit());
+    if let Some(v8) = oj_server::node_compile_cache_opt_in(root) {
+        cmd.env("NODE_COMPILE_CACHE", v8);
+    }
+    let mut child = cmd
         .spawn()
         .map_err(|e| anyhow::anyhow!("could not spawn SSR runner (node): {e}"))?;
 
--- a/crates/oj/src/start_dev.rs
+++ b/crates/oj/src/start_dev.rs
@@ -25,7 +25,10 @@ struct StartState {
     proxy_prefixes: Vec<String>,
     plugin_mw_port: Option<u16>,
     runner: Arc<tokio::sync::Mutex<Runner>>,
-    client_bundle: PathBuf,
+    // The chunk set of the client build this server serves: swapped whole on
+    // watch rebuilds, so every /@oj-start/<chunk> resolves against exactly
+    // one build and a name outside it is a deliberate 404.
+    bundle: std::sync::RwLock<Arc<PinnedBundle>>,
     live_reload: PathBuf,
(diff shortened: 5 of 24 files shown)


raphamorim:
Landed on `main` as 30bd567 (squashed, authored by @aeriksson-lovable). Reviewed the whole design incl. a deep pass on the 1192-line `start_bundle.rs`: the generation key = blake3(salt + each closure file's abs-path + content digest) with the salt folding lockfiles/config/.env, so any graph change edits a closure member's digest → miss (never a false hit); blobs are content-addressed and served through the pinned manifest (chunk names are map keys, never fs paths, so no traversal); `verified_read` refuses mismatched bytes; persist-failure falls back to serving the build dir; prune is LRU, exempts `current`, and is wired at boot (1 GB). `oj_cache` (43 tests) + `e2e/start.mjs` (dev+prod) pass. Thanks André!

PR #17, opened 2026-08-21
persist SSR loader resolutions and load results across boots

aeriksson-lovable:
Part 8 of 16 in a series. Depends on #16 — this PR's diff includes the commits of the PRs below it in the series; to see only this part: https://github.com/aeriksson-lovable/oj/compare/pr/05-client-bundle-cache...pr/06-ssr-loader-caches. Series overview in #12.

---

## Problem

The runner is the Node.js process that executes the app's server code in
start mode. oj spawns it from `crates/oj/src/start_dev.rs`. The runner
registers `crates/oj_server/src/assets/start/loader.mjs` as a module-hooks
loader. The loader resolves every import and transforms every module the
server render touches.

Today the loader recomputes everything on every boot:

1. `resolve()` runs probe chains for aliases, `#imports`, tsconfig paths,
   and relative specifiers, then falls back to Node's resolution. On our
   test app (described under Performance) the first render resolves
   ~44,276 (parentURL, specifier) pairs.
2. `load()` sends every user file through the plugin container. The plugin
   container evaluates the app's Vite config and runs its plugins' hooks
   for the loader. Each call is a round trip. The first render loads ~5,970 modules.
3. `load()` also rebuilt the `import.meta.env` define object on every call
   by scanning `process.env`.

A CPU profile of a warm boot shows the actual `transformSync` work is only
0.05 s. The cost is the resolution probe chains and the per-module round
trips, and both produce the same results every boot when nothing changed.

## Change

Two persistent caches inside the loader, plus one hoist, plus an opt-in V8
compile cache. `OJ_SSR_LOADER_CACHE=off` disables both loader caches.

Both caches are salted with an epoch hash. The epoch is a sha256 over:

- a cache version string and the Node version,
- the app's lockfiles and `package.json`,
- the `vite.config.*` and `oj.config.*` files,
- the five loader asset files themselves (they change with the oj version),
- the serialized define object (it includes every `VITE_*` env var) and
  `TSS_SERVER_FN_BASE`.

If the epoch cannot be computed, or the kill switch is set, both caches
stay off and the loader behaves as before this change.

**Resolution cache.** Maps (parentURL, specifier) to the resolved URL.

1. The cache lives in one JSONL file: `.oj-cache/ssr-resolve.jsonl`. The
   first line carries the epoch. The loader reads the file at init and
   loads the entries into a `Map` when the epoch matches. When the epoch
   does not match, the loader ignores the file and rewrites it on the
   first flush.
2. `resolve()` builds the key from `parentURL` and the specifier. A hit
   with a `file:` URL makes one existence check on the target. If the
   target is gone, the loader deletes the entry and resolves from scratch.
   A hit with a non-`file:` URL (for example `node:` builtins) returns
   directly.
3. Some resolve branches append a version query, `?ojv=<n>`. The runner
   increments `<n>` on every hot reload to bust Node's module registry.
   The cache stores these URLs clean (query removed) with a flag, and a
   hit re-applies the current version. Reloads therefore still bust.
4. A miss runs the original resolution. The loader remembers the result
   unless the URL uses the `ojvirtual:///` scheme (see Correctness). New
   entries queue and append to the JSONL file in batches, off the resolve
   path.

**Load-result cache.** Stores the final transformed source per module.

1. Entries live in `.oj-cache/<2 hex>/<key>.json`. This is the same
   on-disk layout the Rust-side persistent cache uses, so one prune covers
   both. The key is a sha256 over (epoch, branch name, path, input source).
2. For `.ts`/`.tsx` user files, `load()` reads the disk bytes first and
   looks up the entry keyed on those bytes. A hit returns without calling
   the plugin container at all. On a miss, `load()` calls the container.
   When a plugin claims the file (the container returns source that
   differs from the disk bytes), the lookup and the store use a key over
   the plugin output instead. Plugin-claimed files therefore keep the
   container round trip on every boot, and a stale plugin result can never
   be served from a disk-bytes key.
3. `.mdx` files and the CommonJS facades for `node_modules` files get the
   same treatment, keyed on disk bytes under their own branch names. The
   facade case matters because building a facade `require()`s the module
   on the hooks thread only to enumerate export names.
4. Modules whose source contains `import.meta.glob` are never stored. Glob
   expansion depends on directory contents, not on the source bytes, so
   the key would not change when a file is added to a globbed directory.
5. Writes queue and drain off the load path (temp file, then rename). A
   synchronous write per miss would serialize thousands of fs calls into
   first-boot module loading.
6. After the first rendered response, the runner asks the loader for one
   summary line on stderr: hit and miss counts for both caches.

**Hoist.** `viteEnvDefine({ ssr: true })` moves to a module constant. The
epoch hashes the constant, so a define change invalidates both caches.

**Opt-in V8 compile cache.** `OJ_V8_COMPILE_CACHE=on` enables Node's
on-disk compile cache under `.oj-cache/v8` in the runner and the css host,
with explicit `flushCompileCache()` calls, because oj kills its Node
services before Node's exit-time auto-flush would run. This is opt-in
because it measured as a net cost on the test app (see Performance).

**Files changed**

| Path | Change |
| --- | --- |
| `crates/oj_server/src/assets/start/loader.mjs` | Epoch hash, resolution cache, load-result cache, hoisted define constant, stats line. |
| `crates/oj_server/src/assets/start/runner.mjs` | Opt-in V8 compile cache with explicit flushes; requests the stats line after the first response. |
| `crates/oj_server/src/assets/start/css-host.mjs` | Opt-in V8 compile cache with a flush after each compile. |

## Correctness

**Invalidation.** The epoch changes when dependencies, configs, the oj
version, the Node version, or the define/env inputs change. An epoch
change abandons every entry from both caches. Within one epoch, the keys
are content hashes: editing a file changes its disk-bytes key, so only
that module misses and re-transforms.

**Failure handling.** Every cache read and write sits in a try/catch and
degrades to a miss. A load entry that fails to parse deletes itself. A
torn trailing line in the JSONL file (for example after a kill during an
append) fails to parse and is skipped; the other lines still load. (A
later pack-integrity PR frames each JSONL line with a length + payload-hash
prefix and a format-version header, turning this skip into an explicit
truncate-to-last-good-record.) Entry
writes go to a temp file first and rename into place.

**What is never cached, and why.**

- Virtual modules (`ojvirtual:///` URLs, `virtual:` and `\0` specifiers).
  Their resolution and content come from plugin state, not from any file,
  so no file-based key can validate them.
- `import.meta.glob` modules, as described under Change.
- Plugin `load()` output as such: plugin-claimed files keep their
  container round trip every boot, and only the downstream transform is
  cached, keyed on the plugin's current output.

**Hot reload.** Version-query results are stored clean and re-versioned on
every hit with the runner's current version counter. A reload after an
edit therefore gets fresh `?ojv=` URLs exactly as before this change.

**Accepted staleness edge (intentional, documented in the code).** A
resolution hit only re-checks that the cached target still exists. A newly
added file that should shadow an old resolution — for example a new file
that an earlier tsconfig-path pattern now matches — takes effect on the
next epoch change, not instantly. Deleting the resolved file is detected;
adding a better candidate is not.

**Kill switch.** `OJ_SSR_LOADER_CACHE=off` disables both loader caches.
`OJ_V8_COMPILE_CACHE` is off unless set to `on`.

## Performance

concurrently.

Against a stock 0.0.45 control binary run in the same session, this change
alone:

- Warm TTFC: 12 s → 9–10 s. Cold TTFC: 12 s → 12 s (populating the caches
  is free).
- Warm hit rates on the first render: load 5,969/5,970, resolve
  44,276/44,276.
- Edit test: edit one imported `.tsx` file, then boot warm. The loader
  re-transforms only the edited module and renders in 9 s (one run). A
  warm boot after reverting the edit also renders in 9 s (one run).

**Negative result: V8 compile cache.** We measured Node's on-disk compile
cache in the runner on this app, in two A/B comparisons under the same
conditions. It measured as a net cost of 1–3 s per boot: the runner reads,
deserializes, and produces bytecode for ~6k modules. The feature therefore
ships strictly opt-in, via `OJ_V8_COMPILE_CACHE=on`, instead of on by
default.

## How to test

1. Build oj and run `oj dev` in a TanStack Start app. Wait for the first
   `GET /` to return 200. Stop the server.
2. Run `oj dev` again. After the first render, stderr shows one line:
   `oj: ssr loader cache: load <hits>/<total> hits (...), resolve
   <hits>/<total> hits`. Expect near-total hit rates.
3. Inspect `.oj-cache/ssr-resolve.jsonl` (epoch header on line one) and
   the `.oj-cache/<2 hex>/` entry files.
4. Edit one route file. Run `oj dev` again. The load hit count drops by
   the affected modules only, and the page shows the edit.
5. While the server runs, edit a file and confirm the browser shows the
   change. This exercises the `?ojv=` re-versioning path.
6. Run `OJ_SSR_LOADER_CACHE=off oj dev`. No cache line prints and boot
   behavior matches stock.
7. Optional: run `OJ_V8_COMPILE_CACHE=on oj dev` twice and compare boot
   times against the default.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1794,6 +1794,7 @@ dependencies = [
  "axum",
  "clap",
  "notify",
+ "oj_cache",
  "oj_compiler",
  "oj_config",
  "oj_css",
--- a/crates/oj/Cargo.toml
+++ b/crates/oj/Cargo.toml
@@ -10,6 +10,7 @@ name = "oj"
 path = "src/main.rs"
 
 [dependencies]
+oj_cache.workspace = true
 oj_compiler.workspace = true
 oj_env.workspace = true
 oj_config.workspace = true
--- a/crates/oj/src/build.rs
+++ b/crates/oj/src/build.rs
@@ -583,6 +583,7 @@ fn expand_css_via_sidecar(root: &Path, css_file: &Path) -> anyhow::Result<String
             css_file.to_str().unwrap(),
             root.to_str().unwrap(),
         ])
+        .env("NODE_COMPILE_CACHE", oj_server::node_compile_cache(root))
         .current_dir(root)
         .output()
         .context("node not found for tailwind/postcss build")?;
@@ -614,6 +615,7 @@ fn svelte_via_sidecar(root: &Path, file: &Path) -> anyhow::Result<String> {
     .to_string();
     let mut child = std::process::Command::new("node")
         .arg(&script)
+        .env("NODE_COMPILE_CACHE", oj_server::node_compile_cache(root))
         .current_dir(root)
         .stdin(std::process::Stdio::piped())
         .stdout(std::process::Stdio::piped())
@@ -658,6 +660,7 @@ fn preprocess_via_sidecar(root: &Path, css_file: &Path) -> anyhow::Result<String
     .to_string();
     let mut child = std::process::Command::new("node")
         .arg(&script)
+        .env("NODE_COMPILE_CACHE", oj_server::node_compile_cache(root))
         .current_dir(root)
         .stdin(std::process::Stdio::piped())
         .stdout(std::process::Stdio::piped())
@@ -1974,6 +1977,7 @@ pub(crate) async fn build_ssr_app(
         let out = std::process::Command::new("node")
             .arg(&script_path)
             .arg(serde_json::to_string(&paths)?)
+            .env("NODE_COMPILE_CACHE", oj_server::node_compile_cache(root))
             .current_dir(out_dir)
             .output()
             .context("node not found for prerender")?;
--- a/crates/oj/src/ssr_dev.rs
+++ b/crates/oj/src/ssr_dev.rs
@@ -86,17 +86,21 @@ async fn spawn_runner(root: &Path, base: &str, entry_abs: &Path) -> anyhow::Resu
     let script = dir.join("runner.mjs");
     std::fs::write(&script, oj_server::SSR_RUNNER_JS)?;
 
-    let mut child = tokio::process::Command::new("node")
-        .args([
-            "--experimental-vm-modules",
-            &script.to_string_lossy(),
-            base,
-            &entry_abs.to_string_lossy(),
-        ])
-        .current_dir(root)
-        .stdin(Stdio::piped())
-        .stdout(Stdio::piped())
-        .stderr(Stdio::inherit())
+    let mut cmd = tokio::process::Command::new("node");
+    cmd.args([
+        "--experimental-vm-modules",
+        &script.to_string_lossy(),
+        base,
+        &entry_abs.to_string_lossy(),
+    ])
+    .current_dir(root)
+    .stdin(Stdio::piped())
+    .stdout(Stdio::piped())
+    .stderr(Stdio::inherit());
+    if let Some(v8) = oj_server::node_compile_cache_opt_in(root) {
+        cmd.env("NODE_COMPILE_CACHE", v8);
+    }
+    let mut child = cmd
         .spawn()
         .map_err(|e| anyhow::anyhow!("could not spawn SSR runner (node): {e}"))?;
 
--- a/crates/oj/src/start_dev.rs
+++ b/crates/oj/src/start_dev.rs
@@ -25,7 +25,10 @@ struct StartState {
     proxy_prefixes: Vec<String>,
     plugin_mw_port: Option<u16>,
     runner: Arc<tokio::sync::Mutex<Runner>>,
-    client_bundle: PathBuf,
+    // The chunk set of the client build this server serves: swapped whole on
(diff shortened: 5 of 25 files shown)


raphamorim:
Landed on `main` as f1af131 (squashed, authored by @aeriksson-lovable). Reviewed both caches: EPOCH salts over lockfiles/config/loader-assets/node-version/define, so any of those busts the whole cache; the load cache keys plugin-claimed files on the plugin output (never serving a stale plugin result from a disk key) and skips `import.meta.glob` modules; writes are atomic tmp+rename drained off the load path; everything degrades to a miss on error, with `OJ_SSR_LOADER_CACHE=off` as a kill switch. The documented resolve "accepted edge" (a hit only re-checks target existence, so a newly-shadowing file waits for an epoch change) is a reasonable opt-out-able dev tradeoff. `loader-util`/`runner-protocol` unit tests (19) + `e2e/start.mjs` pass. Thanks André!

PR #18, opened 2026-08-21
spawn the start runner concurrently with the client bundle

aeriksson-lovable:
Part 9 of 16 in a series. Depends on #17 — this PR's diff includes the commits of the PRs below it in the series; to see only this part: https://github.com/aeriksson-lovable/oj/compare/pr/06-ssr-loader-caches...pr/07-early-runner. Series overview in #12.

---

## Problem

Definitions used below:

- The runner is the Node.js process that executes the app's server code in
  start mode. oj spawns it from `crates/oj/src/start_dev.rs`
  (`spawn_start_runner`).
- The client bundle is the rolldown build of the browser entry. oj runs it in
  `crates/oj_server/src/assets/start/bundle-client.mjs`.
- The manifest is the module behind the specifier `tanstack-start-manifest:v`.
  `@tanstack/start-server-core` imports it. Its `tsrStartManifest()` returns
  the stylesheet URLs and scripts for the document head.

Today `start_dev` boots in this order:

1. Generate the route tree and the server-function resolver.
2. Run the client bundle and the internal dev server setup, in parallel.
3. Wait for both.
4. Spawn the runner.
5. Bind the listener.

The runner starts last, so its SSR module-graph walk cannot overlap the
client bundle. That walk dominates boot time on large apps.

The runner waits on the client bundle for one reason only.
`crates/oj_server/src/assets/start/loader.mjs` aliases
`tanstack-start-manifest:v` to `manifest.ts` in the start cache directory.
`bundle-client.mjs` overwrites that file when the bundle finishes. It bakes
the collected stylesheet URLs into the file as a literal. The runner imports
the manifest once, during its module-graph walk. If the runner boots before
the bundle finishes, it imports a manifest with no stylesheet URLs and keeps
it. Every served document head then misses its stylesheets.

Measured cost on an 18,600-module app (see Performance): the runner is ready
at +6.9 s after `oj dev` start. The first `GET /` returns 200 at +12.1 s.

## Change

Make the manifest read the stylesheet URLs at call time instead of import
time. That removes the runner's only dependency on the client bundle, so the
runner can spawn before the bundle finishes.

1. A new asset, `crates/oj_server/src/assets/start/manifest-dev.ts`, exports
   `tsrStartManifest()`. On every call it reads `css-urls.json` from the
   start cache directory. A failed read returns an empty list.
2. `loader.mjs` aliases `tanstack-start-manifest:v` to `manifest-dev.ts`.
   The production build path (`build.mjs`) keeps the baked `manifest.ts`.
3. `bundle-client.mjs` writes `css-urls.json` when the bundle finishes. It
   writes a temporary file and renames it, so a concurrent read never sees
   partial JSON.
4. `start_dev.rs` spawns the runner directly after the route tree and the
   server-function resolver complete. The runner now boots concurrently with
   the client bundle and the dev-server setup.
5. `start_dev.rs` then queues one synthetic `GET /` through the runner and
   discards the response. This prewarm render evaluates the SSR module graph
   before the listener binds, so the first real request finds it warm.
6. `crates/oj_server/src/assets/start/runner.mjs` sets `TSS_DEV_SERVER=true`.
   Vite dev sets the same variable. With it, `start-server-core` resolves the
   manifest on every request instead of caching the first result. The
   prewarm render runs before `css-urls.json` exists, so its head has no
   stylesheets; this setting stops that result from being cached. The runner
   also sets `TSS_DEV_SSR_STYLES_ENABLED=false`, because oj serves no
   dev-styles endpoint for that `<link>` injection to hit.
7. `crates/oj_server/src/assets/start/vite-plugin-bridge.mjs` wraps the
   user-plugin `buildStart` loop in a cross-process lock. Both the runner's
   loader and the client bundle run the user plugins' `buildStart` hooks.
   Before this change, boot order serialized the two runs. Now they overlap,
   and plugins that write shared on-disk state must not interleave. The lock:
   `mkdir` of `buildstart.lock` is the atomic acquire; the holder writes its
   pid into the lock directory; a waiter polls every 50 ms; a waiter removes
   the lock if the holder pid is dead, or if the lock is older than 300 s.

Files changed:

| Path | Change |
| --- | --- |
| `crates/oj/src/start_dev.rs` | Spawn the runner after the route tree and the resolver, before the bundle finishes. Queue the discarded prewarm render. `forward` now takes the runner handle instead of the whole server state. |
| `crates/oj_server/src/assets/start/manifest-dev.ts` | New. Lazy manifest: reads `css-urls.json` on every `tsrStartManifest()` call. |
| `crates/oj_server/src/assets/start/loader.mjs` | Alias `tanstack-start-manifest:v` to `manifest-dev.ts` for the runner. |
| `crates/oj_server/src/assets/start/bundle-client.mjs` | Write `css-urls.json` (write-then-rename) when the bundle finishes. |
| `crates/oj_server/src/assets/start/runner.mjs` | Set `TSS_DEV_SERVER=true` and `TSS_DEV_SSR_STYLES_ENABLED=false` (both with `??=`, so a user-set value wins). |
| `crates/oj_server/src/assets/start/vite-plugin-bridge.mjs` | Cross-process lock around the user-plugin `buildStart` loop. |
| `crates/oj_server/src/lib.rs` | Register `manifest-dev.ts` in the written start assets. |

## Correctness

A real request never sees an empty stylesheet list. The listener binds only
after the bundle completes (`tokio::join!` on the bundle precedes the bind),
and `bundle-client.mjs` writes `css-urls.json` before it exits. Only the
prewarm render can read a missing file, and oj discards its response.

`TSS_DEV_SERVER=true` makes `start-server-core` resolve the manifest on
every request. The prewarm render's stylesheet-less head therefore cannot
persist into any later response.

The write-then-rename in `bundle-client.mjs` is atomic on POSIX. A reader
sees either no file or complete JSON, never a partial write.

The prewarm render holds the runner mutex while it runs. A real first
request queues behind it. The worst case is a wait, never a wrong response.

The `buildStart` lock releases in a `finally` block. If the holder process
crashes, a waiter detects the dead pid with `process.kill(pid, 0)` and
removes the lock. If the holder file is unreadable, the 300 s mtime expiry
removes the lock. Every lock filesystem call is wrapped, so a failed cleanup
degrades to the expiry path, not a hang. The lock only serializes the
`buildStart` bodies. It does not change which hooks run, their arguments, or
their order inside one process.

Verified on the test app: the served head's stylesheet set is identical with
and without this change.

Intentionally not covered: the lock does not fix which process runs
`buildStart` first. Before this change, the client bundle always ran first,
because the runner spawned later. A plugin that depends on that cross-process
order is not protected. We did not find such a plugin among the 53 on the
test app.

## Performance

(M-series), release build of oj. TTFC means the wall-clock time from `oj dev`
start until `GET /` returns HTTP 200 with HTML. Warm means the `.oj-cache`
directory is populated from a previous run. Cold means the cache directory is
empty. All measurements ran serially, never concurrently.

With this change applied alone, against a stock 0.0.45 control run in the
same session:

- Runner ready: +6.9 s → +2.8 s after `oj dev` start.
- First `GET /` returning 200: +12.1 s → +8.7 s.
- TTFC: 13 s → 9–10 s.
- Stylesheet check: the served head's stylesheet set is identical with and
  without the change.

Caveat: the cache state (cold or warm) was not recorded per run for the
numbers above.

## How to test

1. Run `oj dev` on a TanStack Start app that imports CSS.
2. Open `view-source:http://localhost:<port>/`. Record every
   `<link rel="stylesheet">` in the head.
3. Run the same app on stock oj. The stylesheet sets must be equal.
4. Measure boot: run `time curl -s -o /dev/null -w '%{http_code}' --retry 60
   --retry-connrefused --retry-delay 1 http://localhost:<port>/` in a second
   shell, started together with `oj dev`. Compare against stock oj.
5. Test the lock steal: before boot, create
   `.oj-cache/start/buildstart.lock/` and write a dead pid into
   `.oj-cache/start/buildstart.lock/holder`. Boot must proceed and remove
   the stale lock.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1794,6 +1794,7 @@ dependencies = [
  "axum",
  "clap",
  "notify",
+ "oj_cache",
  "oj_compiler",
  "oj_config",
  "oj_css",
--- a/crates/oj/Cargo.toml
+++ b/crates/oj/Cargo.toml
@@ -10,6 +10,7 @@ name = "oj"
 path = "src/main.rs"
 
 [dependencies]
+oj_cache.workspace = true
 oj_compiler.workspace = true
 oj_env.workspace = true
 oj_config.workspace = true
--- a/crates/oj/src/build.rs
+++ b/crates/oj/src/build.rs
@@ -583,6 +583,7 @@ fn expand_css_via_sidecar(root: &Path, css_file: &Path) -> anyhow::Result<String
             css_file.to_str().unwrap(),
             root.to_str().unwrap(),
         ])
+        .env("NODE_COMPILE_CACHE", oj_server::node_compile_cache(root))
         .current_dir(root)
         .output()
         .context("node not found for tailwind/postcss build")?;
@@ -614,6 +615,7 @@ fn svelte_via_sidecar(root: &Path, file: &Path) -> anyhow::Result<String> {
     .to_string();
     let mut child = std::process::Command::new("node")
         .arg(&script)
+        .env("NODE_COMPILE_CACHE", oj_server::node_compile_cache(root))
         .current_dir(root)
         .stdin(std::process::Stdio::piped())
         .stdout(std::process::Stdio::piped())
@@ -658,6 +660,7 @@ fn preprocess_via_sidecar(root: &Path, css_file: &Path) -> anyhow::Result<String
     .to_string();
     let mut child = std::process::Command::new("node")
         .arg(&script)
+        .env("NODE_COMPILE_CACHE", oj_server::node_compile_cache(root))
         .current_dir(root)
         .stdin(std::process::Stdio::piped())
         .stdout(std::process::Stdio::piped())
@@ -1974,6 +1977,7 @@ pub(crate) async fn build_ssr_app(
         let out = std::process::Command::new("node")
             .arg(&script_path)
             .arg(serde_json::to_string(&paths)?)
+            .env("NODE_COMPILE_CACHE", oj_server::node_compile_cache(root))
             .current_dir(out_dir)
             .output()
             .context("node not found for prerender")?;
--- a/crates/oj/src/ssr_dev.rs
+++ b/crates/oj/src/ssr_dev.rs
@@ -86,17 +86,21 @@ async fn spawn_runner(root: &Path, base: &str, entry_abs: &Path) -> anyhow::Resu
     let script = dir.join("runner.mjs");
     std::fs::write(&script, oj_server::SSR_RUNNER_JS)?;
 
-    let mut child = tokio::process::Command::new("node")
-        .args([
-            "--experimental-vm-modules",
-            &script.to_string_lossy(),
-            base,
-            &entry_abs.to_string_lossy(),
-        ])
-        .current_dir(root)
-        .stdin(Stdio::piped())
-        .stdout(Stdio::piped())
-        .stderr(Stdio::inherit())
+    let mut cmd = tokio::process::Command::new("node");
+    cmd.args([
+        "--experimental-vm-modules",
+        &script.to_string_lossy(),
+        base,
+        &entry_abs.to_string_lossy(),
+    ])
+    .current_dir(root)
+    .stdin(Stdio::piped())
+    .stdout(Stdio::piped())
+    .stderr(Stdio::inherit());
+    if let Some(v8) = oj_server::node_compile_cache_opt_in(root) {
+        cmd.env("NODE_COMPILE_CACHE", v8);
+    }
+    let mut child = cmd
         .spawn()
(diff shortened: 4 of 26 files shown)


raphamorim:
Landed on `main` as e9c33f8 (squashed, authored by @aeriksson-lovable). The runner now spawns concurrently because the manifest reads `css-urls.json` lazily per-render (`manifest-dev.ts`); the prewarm render is discarded and its CSS-less head never caches (`TSS_DEV_SERVER=true`), and since the listener only binds after the bundle join, real requests always see the CSS. The newly-concurrent `buildStart` is guarded by a cross-process mkdir lock with dead-holder steal + 300s expiry. `e2e/start.mjs` (incl. css-in-head) + oj unit tests pass. Thanks André!

PR #19, opened 2026-08-21
run SSR loader hooks in-thread via module.registerHooks

aeriksson-lovable:
Part 10 of 16 in a series. Depends on #18 — this PR's diff includes the commits of the PRs below it in the series; to see only this part: https://github.com/aeriksson-lovable/oj/compare/pr/07-early-runner...pr/08-inthread-hooks. Series overview in #12.

---

## Problem

oj runs TanStack Start apps in dev mode with a persistent runner. The
runner is the Node.js process that executes the app's server code in start
mode. It lives
in `crates/oj_server/src/assets/start/runner.mjs`. The runner installs a
custom ESM loader, `crates/oj_server/src/assets/start/loader.mjs`. The
loader resolves framework aliases, compiles TS/JSX, and consults the app's
Vite plugins.

Today the runner installs the loader with `module.register()`. Node runs
`module.register()` hooks on a dedicated worker thread. Every `resolve` and
`load` call from the main thread becomes a synchronous cross-thread round
trip through Node's internal `makeSyncRequest`.

An earlier PR in this series made most of these calls cache hits. The
round trip cost remains. A CPU profile of a warm boot showed the main
thread blocked for 2.64 s of a 4.17 s boot inside `makeSyncRequest`
(warm boot, 18,600-module app, macOS arm64, all earlier PRs in this series
applied). Most of those round trips only delivered cache hits. Actual
module execution took about 0.35 s. The thread hop is the dominant cost,
not the work.

Two secondary costs come from the same design:

1. The loader stores one cache file per loaded module. A warm boot opens,
   reads, and parses about 6,000 small JSON files to serve its load hits.
2. The loader evaluates the Vite config inside the hooks worker. Config
   evaluation can mutate `process.env`. Example: a config derives
   `VITE_GIT_COMMIT_SHA` from git state and the loader inlines it into
   `import.meta.env`. Every boot pays for config evaluation before the
   first module loads.

## Change

The runner now installs the loader with `module.registerHooks()`
(Node ≥ 22.15.0 or ≥ 23.5.0). These hooks run synchronously on the main
thread. The hooks worker and its IPC bridge are deleted. A cache hit is
now a `Map` lookup, not a thread hop.

Synchronous hooks cannot `await`. The Vite plugin container is async. The
container therefore moves to a small side worker:

1. `runner.mjs` imports `loader.mjs` directly, then calls
   `module.registerHooks({ resolve, load })`. The loader's own dependency
   graph loads before registration, so the hooks never see it.
2. `loader.mjs` starts the plugin container through a new synchronous
   facade, `crates/oj_server/src/assets/start/container-bridge.mjs`. The
   facade spawns `crates/oj_server/src/assets/start/container-host.mjs` as
   a worker thread and returns call stubs immediately.
3. `container-host.mjs` evaluates the Vite config, builds the plugin
   container, and runs `buildStart`. This all happens off the main thread,
   concurrent with module loading.
4. When a hook needs a plugin answer (a `virtual:` id, an unclaimed user
   file, `.mdx`, `.svg`), the facade posts a request on a `MessageChannel`
   and blocks on `Atomics.wait` until the worker replies. Requests queue in
   the worker until bootstrap finishes, so plugins still observe
   `buildStart` before any `load`.
5. Cache hits never touch the facade. A fully warm boot therefore never
   waits for config evaluation or `buildStart`.

Three behavior differences of `registerHooks` needed explicit handling:

- Sync hooks must return values, not Promises. Every `async`/`await` in the
  hook path is removed; plugin calls go through the blocking facade.
- Sync hooks also intercept `require()` calls; `module.register()` hooks
  did not. The hooks detect `require` traffic through
  `context.conditions.includes("require")` and pass it to `next`
  unmodified. This preserves the old semantics. It also avoids recursion:
  the loader's CJS facade sources `require()` the module they wrap.
- Sync hooks starve the event loop during boot and render, so the loader's
  timer-driven cache writes never fire until the process idles. The loader
  now exports `flushCaches()`. The runner calls it when it signals ready
  and after each response. The per-entry cache drain also writes in
  parallel batches of 64 files instead of one at a time.

The two secondary costs are fixed on top of this structure:

- Load pack. The loader appends every load-cache entry to one JSONL file,
  `.oj-cache/ssr-load-pack.jsonl`. The header line carries the cache epoch
  (the input hash defined by the preceding PR). At init the loader reads
  the file once into a `Map`. A warm boot serves its ~6,000 load hits from
  that `Map` instead of ~6,000 per-file open/read/parse round trips. The
  pack is an index over the per-entry cache, not a replacement. A key
  missing from the pack falls back to the per-file path, and the loader
  re-appends such hits, so the pack converges toward complete coverage.
- Env delta. `container-host.mjs` records which `process.env` keys config
  evaluation changed, and reports that delta on request. The loader merges
  the delta into the `import.meta.env` define source, so values such as
  `VITE_GIT_COMMIT_SHA` inline correctly even though config evaluation now
  runs in the worker's env copy. The loader persists the delta to
  `.oj-cache/ssr-env.json`, keyed by a hash of every input that can change
  it: config files, lockfiles, the loader assets, `.env*` files, the git
  `HEAD` state (worktree-aware, loose or packed refs), and the Node
  version. On a key hit the loader uses the stored delta and does not
  block on the worker at all. On a key change the boot recompiles anyway,
  so blocking once there costs little.

Files changed:

| Path | Change |
| --- | --- |
| `crates/oj_server/src/assets/start/runner.mjs` | Install hooks via `module.registerHooks()`; drop the `register()` `MessageChannel`; call `flushCaches()` at ready and after each response. |
| `crates/oj_server/src/assets/start/loader.mjs` | Hooks become synchronous; add the load pack; add the persisted env delta; batch the per-entry cache drain; export `flushCaches`, `setVersion`, `reportCacheStats`. |
| `crates/oj_server/src/assets/start/container-bridge.mjs` | New. Synchronous facade over the plugin container: worker spawn, `Atomics.wait` request loop, 300 s deadline. |
| `crates/oj_server/src/assets/start/container-host.mjs` | New. Worker that owns the async plugin container and reports the config-eval env delta. |
| `crates/oj_server/src/assets/start/vite-plugin-bridge.mjs` | Export `findConfig` so the facade can skip spawning a worker for apps with no Vite config. |
| `crates/oj_server/src/assets/start/resolve-pkg.mjs` | `viteEnvDefine` accepts an env source object instead of reading `process.env` directly. |
| `crates/oj_server/src/lib.rs` | Embed the two new assets. |
| `crates/oj_server/src/assets/start/ARCHITECTURE.md` | Describe the new loader/runner/container topology. |

## Correctness

- Node version requirement: `module.registerHooks()` exists in Node
  ≥ 22.15.0 and ≥ 23.5.0. This PR raises the minimum Node version for the
  start-mode dev runner to those releases. There is no fallback to
  `module.register()`. On older Node the runner fails at startup with a
  clear `registerHooks is not a function` error. We left a fallback path
  out because it doubles the loader entry surface; we can add one if you
  prefer it.
- Hook ordering: the worker queues plugin requests until bootstrap
  (config eval and `buildStart`) completes. Plugins observe the same
  "buildStart before any load" ordering as before.
- `require()` traffic passes through `next` untouched, so CommonJS
  resolution behaves exactly as it did under `module.register()`, which
  never saw it.
- The facade's `Atomics.wait` loop has a 300 s deadline. `buildStart` can
  legitimately block on the cross-process lock in
  `vite-plugin-bridge.mjs`, which another process can hold for up to
  300 s; the deadline matches that horizon. On timeout the loader throws
  instead of hanging.
- The worker and its port are `unref()`ed. They cannot keep the runner
  alive after stdin closes.
- A container failure in the worker logs to stderr and answers `null` to
  every plugin call, matching the previous degrade path.
- Load pack: gated on the same epoch and the same
  `OJ_SSR_LOADER_CACHE=off` kill switch as the per-entry cache. An epoch
  mismatch discards the pack. A corrupt line is skipped; the entry then
  falls back to the per-file cache or a recompute.
- Env delta: `OJ_SSR_LOADER_CACHE=off` also disables persistence; the
  loader then always fetches the delta fresh from the worker. Known
  compromise: an env side effect that is nondeterministic for identical
  inputs (for example a timestamp) replays from the persisted delta on
  warm boots. The previous design recomputed it every boot. Any input the
  key tracks (config, lockfile, `.env*`, git `HEAD`, Node version, loader
  assets) invalidates the stored delta.
- Not covered: env side effects driven by inputs outside the key (for
  example a network call inside the config). Delete
  `.oj-cache/ssr-env.json` or set `OJ_SSR_LOADER_CACHE=off` to force
  recomputation.

## Performance

concurrently.

Numbers below are measured on top of the earlier PRs in this series, since
this PR restructures the loader those PRs introduced.

- Before: a CPU profile of a warm boot showed the main thread blocked
  2.64 s of 4.17 s in `makeSyncRequest`, mostly delivering cache hits.
  Module execution was about 0.35 s.
- After: warm TTFC 4 s → 3 s. Cold TTFC 9 s → 7 s.
- The cold improvement comes from overlap: container bootstrap (config
  eval, `buildStart`) now runs concurrent with module loading instead of
  ahead of it.

## How to test

1. Use Node ≥ 22.15.0 (or ≥ 23.5.0).
2. Run `oj dev` on a TanStack Start app. Request `/`. The page renders as
   before.
3. Run `oj dev` again (warm). Confirm `.oj-cache/ssr-load-pack.jsonl` and
   `.oj-cache/ssr-env.json` exist. The loader's stats line on stderr
   reports the same hit counts as before this PR.
4. Env delta: add `process.env.VITE_PROBE = "x"` at the top level of the
   app's `vite.config.ts`, and render `import.meta.env.VITE_PROBE`
   somewhere. It renders `"x"` on both cold and warm boots.
5. Invalidation: commit anything in the app repo (git `HEAD` changes),
   boot again, and confirm a config-derived value such as a commit SHA
   updates.
6. Kill switch: `OJ_SSR_LOADER_CACHE=off oj dev` ignores the pack and the
   persisted delta and still boots correctly.
7. CJS interop: an app dependency loaded through `require()` (any CJS
   package) still resolves; the hooks pass `require` traffic through.

A note on part 5 (#27): the in-thread hooks in this PR happened to mask a module-resolution failure introduced by the stdin startup check from part 1 — in-thread loading keeps refreshing the shared stat buffer that `fs.realpathSync` reads, so the stale entry never survives long enough to poison a resolution. #27, below this PR in the series, fixes that failure at its source.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1794,6 +1794,7 @@ dependencies = [
  "axum",
  "clap",
  "notify",
+ "oj_cache",
  "oj_compiler",
  "oj_config",
  "oj_css",
--- a/crates/oj/Cargo.toml
+++ b/crates/oj/Cargo.toml
@@ -10,6 +10,7 @@ name = "oj"
 path = "src/main.rs"
 
 [dependencies]
+oj_cache.workspace = true
 oj_compiler.workspace = true
 oj_env.workspace = true
 oj_config.workspace = true
--- a/crates/oj/src/build.rs
+++ b/crates/oj/src/build.rs
@@ -583,6 +583,7 @@ fn expand_css_via_sidecar(root: &Path, css_file: &Path) -> anyhow::Result<String
             css_file.to_str().unwrap(),
             root.to_str().unwrap(),
         ])
+        .env("NODE_COMPILE_CACHE", oj_server::node_compile_cache(root))
         .current_dir(root)
         .output()
         .context("node not found for tailwind/postcss build")?;
@@ -614,6 +615,7 @@ fn svelte_via_sidecar(root: &Path, file: &Path) -> anyhow::Result<String> {
     .to_string();
     let mut child = std::process::Command::new("node")
         .arg(&script)
+        .env("NODE_COMPILE_CACHE", oj_server::node_compile_cache(root))
         .current_dir(root)
         .stdin(std::process::Stdio::piped())
         .stdout(std::process::Stdio::piped())
@@ -658,6 +660,7 @@ fn preprocess_via_sidecar(root: &Path, css_file: &Path) -> anyhow::Result<String
     .to_string();
     let mut child = std::process::Command::new("node")
         .arg(&script)
+        .env("NODE_COMPILE_CACHE", oj_server::node_compile_cache(root))
         .current_dir(root)
         .stdin(std::process::Stdio::piped())
         .stdout(std::process::Stdio::piped())
@@ -1974,6 +1977,7 @@ pub(crate) async fn build_ssr_app(
         let out = std::process::Command::new("node")
             .arg(&script_path)
             .arg(serde_json::to_string(&paths)?)
+            .env("NODE_COMPILE_CACHE", oj_server::node_compile_cache(root))
             .current_dir(out_dir)
             .output()
             .context("node not found for prerender")?;
--- a/crates/oj/src/ssr_dev.rs
+++ b/crates/oj/src/ssr_dev.rs
@@ -86,17 +86,21 @@ async fn spawn_runner(root: &Path, base: &str, entry_abs: &Path) -> anyhow::Resu
     let script = dir.join("runner.mjs");
     std::fs::write(&script, oj_server::SSR_RUNNER_JS)?;
 
-    let mut child = tokio::process::Command::new("node")
-        .args([
-            "--experimental-vm-modules",
-            &script.to_string_lossy(),
-            base,
-            &entry_abs.to_string_lossy(),
-        ])
-        .current_dir(root)
-        .stdin(Stdio::piped())
-        .stdout(Stdio::piped())
-        .stderr(Stdio::inherit())
+    let mut cmd = tokio::process::Command::new("node");
+    cmd.args([
+        "--experimental-vm-modules",
+        &script.to_string_lossy(),
+        base,
+        &entry_abs.to_string_lossy(),
+    ])
+    .current_dir(root)
+    .stdin(Stdio::piped())
+    .stdout(Stdio::piped())
+    .stderr(Stdio::inherit());
+    if let Some(v8) = oj_server::node_compile_cache_opt_in(root) {
+        cmd.env("NODE_COMPILE_CACHE", v8);
+    }
+    let mut child = cmd
         .spawn()
         .map_err(|e| anyhow::anyhow!("could not spawn SSR runner (node): {e}"))?;
 
--- a/crates/oj/src/start_dev.rs
+++ b/crates/oj/src/start_dev.rs
@@ -72,11 +72,24 @@ pub async fn start_dev(
         config: None,
     }
     .build_app();
+    resolver.await??;
+    // The runner's SSR module-graph walk dominates boot time; start it now so
+    // it overlaps the client bundle (the manifest reads CSS lazily, so the
+    // runner no longer needs the bundle's output).
+    let runner = Arc::new(tokio::sync::Mutex::new(
+        spawn_start_runner(&root, &cache).await?,
+    ));
+    {
+        // Prewarm: queue one synthetic render so first-request module
+        // evaluation also happens before the listener binds.
+        let runner = Arc::clone(&runner);
+        tokio::spawn(async move {
+            let _ = forward(&runner, "GET".into(), "/".into(), vec![], None).await;
+        });
+    }
     let (bundle_res, built_res) = tokio::join!(bundle, built_fut);
     let pinned = bundle_res??;
-    resolver.await??;
     let built = built_res?;
-    let runner = spawn_start_runner(&root, &cache).await?;
     let css_host = if app_uses_tailwind(&root) {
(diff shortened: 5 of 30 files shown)


raphamorim:
Landed on `main` as 6bba2e9 (squashed from the 3 branch commits, authored by @aeriksson-lovable). Fork-reviewed the Atomics sync facade in depth: the 4-byte signal SAB + MessagePort payload with a 200ms bounded `Atomics.wait` poll can't deadlock or miss a wakeup, requests are strictly serial with stale-reply skip, and a worker crash hits the 300s deadline rather than hanging; `require()` pass-through via `context.conditions` avoids cjs-facade recursion; the JSONL pack is a convergent index over the authoritative per-file cache; env-delta never serves wrong-empty. One finding fixed while landing: the runner's loader contract changed (registerHooks + flushCaches/setVersion/reportCacheStats), so I updated `runner-protocol.test.mjs`'s stub loader to match (it still only exported the old `initialize`). `start.mjs` + `oj_server` + loader/runner unit tests (19) all pass. Thanks André!

PR #20, opened 2026-08-21
cache the vite config extract, keyed by the config's import graph

aeriksson-lovable:
Part 11 of 16 in a series. Depends on #19 — this PR's diff includes the commits of the PRs below it in the series; to see only this part: https://github.com/aeriksson-lovable/oj/compare/pr/08-inthread-hooks...pr/09-config-extract-cache. Series overview in #12.

---

## Problem

oj reads four values from the app's Vite config before it can serve: `base`,
`server.port`, `define`, and the plugin list. To read them, oj spawns an
extract subprocess. The extract is a Node.js process that runs
`crates/oj_server/src/assets/vite-extract.mjs`. The script evaluates the
config file with Vite's `loadConfigFromFile` and prints the four values as
JSON. `extract_vite_values` in `crates/oj_server/src/plugins.rs` spawns it.

Evaluating the config file executes all of its imports. On a large app those
imports are expensive. We measured on a large real-world app: a TanStack
Start application with approximately 18,600 client modules, approximately
2,500 SSR modules, and a 788-line Vite config that registers 53 plugins.
Hardware: macOS arm64 (M-series), release build of oj. On this app the config
imports wrangler and miniflare, and the extract takes 0.78–1.24 s per boot.

The extract runs on every boot. Its output is identical on every boot unless
the config or its inputs change. The plugin host (the Node.js process that
runs the app's Vite plugins) cannot spawn until the extract finishes, so the
extract serializes the boot.

There is a second scheduling problem. `start_dev` in
`crates/oj/src/start_dev.rs` ran `build_app` (the step that runs the extract
and spawns the plugin host) after route codegen, the resolver, and the runner
spawn. `build_app` needs only the app root, so this wait is unnecessary.

## Change

1. `vite-extract.mjs` now reports the config's import graph. Vite's
   `loadConfigFromFile` returns a `dependencies` array: every file it bundled
   while evaluating the config. The script resolves these paths against the
   app root and prints them as `__deps`, plus `__ok: true` on success. The
   esbuild fallback path reports the esbuild metafile inputs instead.
2. A new store, `ConfigExtractStore` in
   `crates/oj_cache/src/config_extract.rs`, caches the extract output under
   `.oj-cache/config-extract/`. The cache key (the entry file name) is a
   blake3 hash of: a salt (the oj version plus a blake3 hash of the embedded
   extract script), the config file path, the Vite command (`serve`), and the
   mode (`development`).
3. Each cache entry stores: a blake3 content digest for the config file and
   for every file in `__deps` (58 files on the test app), an env epoch, the
   extract's stdout, and the extract's stderr.
4. The env epoch is one blake3 hash over the non-file inputs that can change
   the extracted values: the lockfiles (`package-lock.json`, `yarn.lock`,
   `pnpm-lock.yaml`, `bun.lockb`), `package.json`, the `.env`, `.env.local`,
   `.env.development`, and `.env.development.local` files, and the process
   environment variables that app configs commonly read (`VITE_*` and a small
   allowlist such as `NODE_ENV` and `CI`).
5. On boot, `extract_vite_values` looks the entry up first. A hit requires:
   the entry parses as JSON, the stored env epoch equals the current env
   epoch, and every stored file digest equals the file's current digest. On a
   hit, oj replays the stored stderr (so config warnings still print) and
   returns the stored values without spawning the extract. Any mismatch is a
   miss, and a miss runs the extract exactly as before.
6. After a successful extract (`__ok: true` in the output), oj writes the
   entry. Writes go to a temporary file first and then rename into place, so
   a killed process cannot leave a half-written entry visible.
7. `start_dev` now spawns `build_app` as its own task at the top of the
   function, before codegen. The extract (or its cache lookup) and the plugin
   host's own config evaluation now overlap route codegen, the resolver, the
   runner spawn, and the client bundle. On a cache hit the plugin host spawns
   at +3 ms after boot start on the test app, instead of after the extract.
8. `OJ_BOOT_PHASES=1` prints epoch-millisecond timestamps at the boot phase
   boundaries (`build_app begin`, `plugin host spawning`, `listening`, cache
   hit/miss, and others) to stderr. This is how we measured this change. It
   is off by default.

Files changed:

| Path | Change |
| --- | --- |
| `crates/oj_cache/src/config_extract.rs` | New. `ConfigExtractStore`: key derivation, lookup with digest validation, atomic store. Unit tests for hit/miss behavior. |
| `crates/oj_cache/src/lib.rs` | Registers the `config_extract` module. |
| `crates/oj_server/src/assets/vite-extract.mjs` | Prints `__ok` and `__deps` (the config's import graph) next to the extracted values. |
| `crates/oj_server/src/plugins.rs` | `extract_vite_values` consults the store before spawning the extract, and writes an entry after a successful run. |
| `crates/oj/src/start_dev.rs` | Spawns `build_app` as a task at the top of `start_dev` and joins it with the client bundle. |
| `crates/oj_server/src/lib.rs` | Adds the `boot_phase` helper behind `OJ_BOOT_PHASES=1`. |

## Correctness

The cache never changes what oj computes. It only skips the subprocess when
every input it tracks is unchanged. The tracked inputs are:

- The extract script itself and the oj version (in the key salt). A new oj
  build or a changed script gets a new key and therefore a cold cache.
- The config file path, the Vite command, and the mode (in the key).
- The content of the config file and of every file in its import graph
  (content digests, validated on every lookup). An edit, a deletion, or a
  new file at a tracked path is a miss.
- The env epoch: lockfiles, `package.json`, `.env*` files, and the allowlisted
  environment variables. A dependency upgrade or a changed `VITE_*` variable
  is a miss.

Failure handling:

- Failures are never cached. oj writes an entry only when the extract prints
  `__ok: true`. A crashed or throwing extract produces no entry, so the next
  boot retries.
- A cache entry that does not parse as JSON is deleted and treated as a miss.
- Every filesystem error in the store (read, create, write, rename) degrades
  to a miss. The extract then runs exactly as without this change.
- Writes are atomic (temporary file plus rename), so concurrent boots cannot
  observe a partial entry.
- Stderr from the original run is stored and replayed on every hit, so
  warnings that the config evaluation prints (for example unsupported-option
  warnings) do not disappear when the cache hits.

Edge cases considered:

- A tracked file that was absent at store time and is still absent at lookup
  time is a hit (absent == absent). If it appears, its digest changes and the
  lookup is a miss.
- Version-skew between entries from different oj builds cannot occur: the
  version is in the key, so old entries are never looked up. They are small
  JSON files and are left in place.

Intentionally not covered:

- Configs that read environment variables outside the allowlist. The epoch
  cannot enumerate arbitrary `process.env` reads. The allowlist covers the
  variables Vite itself and common configs read. A user can delete
  `.oj-cache/config-extract/` to force a re-run.
- The last-resort load path (no Vite and no esbuild resolvable) imports the
  config directly and reports no dependencies. Entries from that path are
  invalidated only by the config file itself and the env epoch, not by files
  the config imports. Vite is present in every app oj currently targets, so
  this path is rare.
- Non-deterministic configs (for example a config that reads the clock). The
  cache replays the first run's values until an input changes. The extract
  already assumed the values are stable across the boot; the cache extends
  that assumption across boots with unchanged inputs.

## Performance

Test app: a TanStack Start application with approximately 18,600 client
modules, approximately 2,500 SSR modules, and a 788-line Vite config that
registers 53 plugins. Hardware: macOS arm64 (M-series), release build of oj.
Warm means the `.oj-cache` directory is populated from a previous run. All
measurements ran serially, never concurrently. Phase times come from the
`OJ_BOOT_PHASES=1` timestamps.

- Extract phase: 1,240 ms (miss, subprocess runs) → 2 ms (warm hit).
- Pre-listen time (from `oj dev` start until oj binds its HTTP listener):
  2.36 s → 1.05 s warm, with the early `build_app` spawn.
- The two numbers above are single measurements; run counts and variance
  were not recorded.

Honest caveat: on this branch alone, TTFC (wall-clock time from `oj dev`
start until `GET /` returns HTTP 200 with HTML) does not improve. The SSR
render path still gates TTFC at this point, so the saved extract time hides
behind it. The saving becomes end-to-end visible in combination with #19
in this series ("run SSR loader hooks in-thread via
module.registerHooks"), which removes that gate. The pre-listen improvement
above is real in isolation.

## How to test

1. Pick an app with a `vite.config.ts`. Run `OJ_BOOT_PHASES=1 oj dev` once.
   stderr shows `vite-extract cache miss (subprocess ran)`.
2. Stop oj. Run it again. stderr shows `vite-extract cache hit`, and the
   `plugin host spawning` timestamp moves close to `build_app begin`. The
   served app behaves identically.
3. Edit `vite.config.ts` (or any file it imports, or a lockfile, or a
   `.env` file, or set a new `VITE_*` variable). The next boot is a miss and
   re-runs the extract.
4. Add a config warning (for example an unsupported option). Boot twice. The
   warning prints on the hit boot too.
5. Corrupt an entry: `echo garbage > .oj-cache/config-extract/<hash>.json`.
   The next boot is a miss and the entry is replaced.
6. Unit tests: `cargo test -p oj_cache config_extract`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1794,6 +1794,7 @@ dependencies = [
  "axum",
  "clap",
  "notify",
+ "oj_cache",
  "oj_compiler",
  "oj_config",
  "oj_css",
--- a/crates/oj/Cargo.toml
+++ b/crates/oj/Cargo.toml
@@ -10,6 +10,7 @@ name = "oj"
 path = "src/main.rs"
 
 [dependencies]
+oj_cache.workspace = true
 oj_compiler.workspace = true
 oj_env.workspace = true
 oj_config.workspace = true
--- a/crates/oj/src/build.rs
+++ b/crates/oj/src/build.rs
@@ -583,6 +583,7 @@ fn expand_css_via_sidecar(root: &Path, css_file: &Path) -> anyhow::Result<String
             css_file.to_str().unwrap(),
             root.to_str().unwrap(),
         ])
+        .env("NODE_COMPILE_CACHE", oj_server::node_compile_cache(root))
         .current_dir(root)
         .output()
         .context("node not found for tailwind/postcss build")?;
@@ -614,6 +615,7 @@ fn svelte_via_sidecar(root: &Path, file: &Path) -> anyhow::Result<String> {
     .to_string();
     let mut child = std::process::Command::new("node")
         .arg(&script)
+        .env("NODE_COMPILE_CACHE", oj_server::node_compile_cache(root))
         .current_dir(root)
         .stdin(std::process::Stdio::piped())
         .stdout(std::process::Stdio::piped())
@@ -658,6 +660,7 @@ fn preprocess_via_sidecar(root: &Path, css_file: &Path) -> anyhow::Result<String
     .to_string();
     let mut child = std::process::Command::new("node")
         .arg(&script)
+        .env("NODE_COMPILE_CACHE", oj_server::node_compile_cache(root))
         .current_dir(root)
         .stdin(std::process::Stdio::piped())
         .stdout(std::process::Stdio::piped())
@@ -1974,6 +1977,7 @@ pub(crate) async fn build_ssr_app(
         let out = std::process::Command::new("node")
             .arg(&script_path)
             .arg(serde_json::to_string(&paths)?)
+            .env("NODE_COMPILE_CACHE", oj_server::node_compile_cache(root))
             .current_dir(out_dir)
             .output()
             .context("node not found for prerender")?;
--- a/crates/oj/src/ssr_dev.rs
+++ b/crates/oj/src/ssr_dev.rs
@@ -86,17 +86,21 @@ async fn spawn_runner(root: &Path, base: &str, entry_abs: &Path) -> anyhow::Resu
     let script = dir.join("runner.mjs");
     std::fs::write(&script, oj_server::SSR_RUNNER_JS)?;
 
-    let mut child = tokio::process::Command::new("node")
-        .args([
-            "--experimental-vm-modules",
-            &script.to_string_lossy(),
-            base,
-            &entry_abs.to_string_lossy(),
-        ])
-        .current_dir(root)
-        .stdin(Stdio::piped())
-        .stdout(Stdio::piped())
-        .stderr(Stdio::inherit())
+    let mut cmd = tokio::process::Command::new("node");
+    cmd.args([
+        "--experimental-vm-modules",
+        &script.to_string_lossy(),
+        base,
+        &entry_abs.to_string_lossy(),
+    ])
+    .current_dir(root)
+    .stdin(Stdio::piped())
+    .stdout(Stdio::piped())
+    .stderr(Stdio::inherit());
+    if let Some(v8) = oj_server::node_compile_cache_opt_in(root) {
+        cmd.env("NODE_COMPILE_CACHE", v8);
+    }
+    let mut child = cmd
         .spawn()
         .map_err(|e| anyhow::anyhow!("could not spawn SSR runner (node): {e}"))?;
 
--- a/crates/oj/src/start_dev.rs
+++ b/crates/oj/src/start_dev.rs
@@ -52,6 +52,20 @@ pub async fn start_dev(
     oj_server::write_start_assets(&cache)?;
     oj_server::boot_phase("start_dev begin");
 
+    // build_app (vite.config extract + plugin host boot) needs only the root;
+    // starting it first overlaps the plugin host's own config evaluation with
+    // route/resolver codegen, the runner spawn, and the client bundle.
+    let built_task = tokio::spawn(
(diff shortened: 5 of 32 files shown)


raphamorim:
Landed on `main` as 72c5490 (squashed from the 3 branch commits, authored by @aeriksson-lovable). Reviewed the cache: key = blake3(oj-version + extract-script + config-path + command/mode), and a hit additionally requires an unchanged `env_epoch` (lockfiles/`.env`/relevant env vars) plus matching content digests of the config file and its **full import graph** (`__deps` from `loadConfigFromFile` / esbuild metafile), so any config-graph or env change re-runs the subprocess; only `__ok` results are stored, corrupt entries self-delete, writes are atomic. The `build_app` hoist overlaps config eval with codegen/bundle, and `OJ_BOOT_PHASES` is off-by-default instrumentation. `oj_cache` (49 tests), `config-function`/`config-flag`, and `start.mjs` pass. Thanks André!

PR #21, opened 2026-08-21
serve SSR load-pack hits by file offset to bound runner memory

aeriksson-lovable:
Part 12 of 16 in a series. Depends on #20 — this PR's diff includes the commits of the PRs below it in the series; to see only this part: https://github.com/aeriksson-lovable/oj/compare/pr/09-config-extract-cache...pr/10-memory-pass. Series overview in #12.

---

Depends on #19 ("run SSR loader hooks in-thread via
module.registerHooks"): that PR introduces the JSONL load pack this PR
changes.

## Problem

Definitions first. The runner is the Node.js process that executes the
app's server code in start mode. oj spawns it from
`crates/oj/src/start_dev.rs`. The loader
(`crates/oj_server/src/assets/start/loader.mjs`) runs inside the runner and
serves every module resolve and load. The load pack is the loader's warm
cache for compiled module sources: one JSONL file at
`.oj-cache/ssr-load-pack.jsonl`. Its header line carries the epoch, a hash
of the config and lockfiles. Each following line is one pack entry:
`{"k":"<64-hex key>","c":"<compiled source>"}`. The per-entry cache (one
JSON file per key under `.oj-cache/`) stays authoritative; the pack only
batches it. (A later pack-integrity PR frames each line with a
length + payload-hash prefix and a format-version header; the offset index
and pread path are unchanged.)

Before this change, the loader read the pack file once at init and stored
every entry's compiled source in a `Map` for the process lifetime. On our
test app this `Map` held about 20 million characters. The `Map` also grew
without bound during a dev session: `rememberPack` added an entry for every
newly compiled module, and no entry was ever removed. Entries that a boot
never serves — modules superseded by edits within one epoch, and routes
that a session never evaluates — stayed in memory forever.

We measured before changing anything. Steady-state RSS per process (warm
boot, 18,600-module app, macOS arm64; conditions in Performance below):

| Process | RSS after first 200 | RSS settled |
| --- | --- | --- |
| runner | 1,096 MB | 746 MB |
| plugin host | 620 MB | 472 MB |
| css host | 41 MB | 41 MB |
| oj (Rust) | 31 MB | 31 MB |
| total | 1,788 MB | — |

Heap snapshots of the runner attributed the pack's 20 million characters.
The result was negative for the obvious suspect: entries that a boot serves
alias the module sources V8 already retains for loaded scripts. Deleting
those entries after they were served freed only about 0.2 MB. The real cost
is the entries that are never served, plus the unbounded growth. So the fix
targets retention and growth, not the served entries.

## Change

The loader now indexes the pack file instead of materializing it.

1. At init, the loader reads the pack file once. For each valid line it
   records only the byte offset and length in a new `packIndex` `Map`. A
   valid line starts with `{`, is longer than 70 bytes, and carries a
   64-hex key at the fixed position bytes 6–70. Invalid lines are skipped.
2. The loader opens the pack file read-only and keeps the file descriptor
   for the process lifetime.
3. A cache hit calls `packGet(key)`. It checks the in-memory `Map` first,
   then the index. An index hit is one positioned `readSync` (a single
   `pread`) from the open descriptor, then one `JSON.parse`. The parsed
   entry must repeat the key and hold a string source; otherwise the loader
   deletes the index entry and reports a miss.
4. A miss falls through to the per-entry cache, unchanged. Such hits are
   re-appended to the pack, so the pack still converges.
5. New entries keep their source in memory only until the next flush.
   `flushLoadPack` appends the batch to the file, records each entry's byte
   offset in `packIndex`, and deletes the in-memory source. Steady state
   therefore holds offsets, not sources.
6. `OJ_SSR_MEM_STATS=1` adds a one-shot diagnostic. After the first
   rendered response, the runner prints one stderr line with its RSS, its
   V8 heap statistics, the loader's own accounting (`packIndexEntries`,
   `packMemEntries`, `packMemChars`, `resolveEntries`, `resolveChars`), and
   the plugin-container worker's heap via a new `__heap` bridge call.
7. A small fix rides along: a stub loader set via `OJ_RUNNER_LOADER` (the
   test `e2e/unit/runner-protocol.test.mjs` uses one) exports none of
   `resolve` / `load` / `flushCaches` / `setVersion` / `reportCacheStats`.
   The runner crashed on the `flushCaches` call before it printed ready.
   All loader API calls are now optional.

Files changed:

| Path | Change |
| --- | --- |
| `crates/oj_server/src/assets/start/loader.mjs` | Pack init indexes byte offsets; `packGet` serves hits with one `pread`; flush moves entries to the index and drops sources; new `memStats()` export. |
| `crates/oj_server/src/assets/start/runner.mjs` | All loader API calls optional (stub-loader fix); `OJ_SSR_MEM_STATS=1` one-shot diagnostic after the first response. |
| `crates/oj_server/src/assets/start/container-bridge.mjs` | New `heap()` bridge method. |
| `crates/oj_server/src/assets/start/container-host.mjs` | `__heap` handler returns `v8.getHeapStatistics()` for the container worker. |

## Correctness

The per-entry cache stays authoritative. Every failure path in `packGet`
degrades to a miss: a failed read, a failed parse, a key mismatch, or a
non-string source deletes the index entry and falls through to the
per-entry cache. A miss that hits the per-entry cache re-appends the entry,
so the pack heals itself.

Edge cases:

- Corrupt or truncated line: skipped at init by the shape check, or
  rejected in `packGet` by the key/type check. Result is a per-entry-cache
  hit, not a failure.
- Epoch mismatch: the whole pack is ignored, as before this change.
- Read-during-append: appends run on a serialized promise chain
  (`packFlushChain`). An entry enters `packIndex` only after its append
  completes, so a recorded offset always points at fully written bytes.
- Fresh epoch: the first flush rewrites the file with a new header. The
  descriptor opens after that write, so it never points at a replaced file.
- Stub loaders: `module.registerHooks` is called only when the loader
  exports `resolve` or `load`; all other loader calls use optional
  chaining. `e2e/unit/runner-protocol.test.mjs` covers this path.

Intentionally not covered: `packIndex` itself still grows during a session
(one key string plus two integers per entry). We accept that; it is the
index, not the sources. The pack file also grows on disk within an epoch;
an epoch change resets it.

We also measured two other caches and deliberately left them alone:

- The loader's resolve cache (`.oj-cache/ssr-resolve.jsonl`, held in a
  `Map`) weighs 15 MB in memory, but every retained entry served a hit
  during the measured session (100% live hit rate). Evicting it would trade
  15 MB for repeated resolution work. Kept as is.
- The Rust `MemoryCache` in `crates/oj_server/src/lib.rs` is empty in start
  mode. No change.

## Performance

concurrently.

Method: warm boots, RSS sampled per process at the first 200 response and
again after the processes settle; heap snapshots of the runner for
attribution.

- Before: total RSS across the four oj processes at the first 200 response
  was 1,788 MB (runner 1,096 MB, settling to 746 MB; plugin host 620 MB,
  settling to 472 MB; css host 41 MB; Rust oj 31 MB).
- After: total RSS settles at 1,265 MB, and the loader's pack no longer
  grows with session length — steady state holds byte offsets only.
- Negative result, reported on purpose: deleting pack entries after they
  were served freed only about 0.2 MB, because those strings alias module
  sources V8 retains anyway. That measurement redirected the fix from
  eviction of served entries to never materializing unserved ones.
- Left alone with numbers: the resolve cache (15 MB, 100% live hit rate)
  and the Rust `MemoryCache` (empty in start mode).
- The per-hit `pread` is TTFC-neutral within run-to-run noise. Control at
  the same base: cold 7 s (1 run), warm 5 s, 3 s, 3 s, 3 s (4 runs). With
  this change, two series: cold 7 s, warm 4 s, 3 s, 3 s; and cold 7 s,
  warm 5 s, 3 s, 3 s (1 cold and 3 warm runs each).

## How to test

1. Build oj and run `oj dev` twice on an app so `.oj-cache` is warm.
2. Run `OJ_SSR_MEM_STATS=1 oj dev`, request `/` once, and read the stderr
   line `oj ssr memstats: {...}`. On a warm boot, expect `packIndexEntries`
   near the module count and `packMemEntries` / `packMemChars` near zero.
3. Delete `.oj-cache/ssr-load-pack.jsonl` and boot again. The boot must
   succeed from the per-entry cache, and the pack file must regenerate.
4. Truncate the last line of `ssr-load-pack.jsonl` mid-entry and boot. The
   damaged entry must degrade to a per-entry-cache hit, not an error.
5. Run `node --test e2e/unit/runner-protocol.test.mjs`. It exercises the
   stub-loader path that previously crashed the runner.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1794,6 +1794,7 @@ dependencies = [
  "axum",
  "clap",
  "notify",
+ "oj_cache",
  "oj_compiler",
  "oj_config",
  "oj_css",
--- a/crates/oj/Cargo.toml
+++ b/crates/oj/Cargo.toml
@@ -10,6 +10,7 @@ name = "oj"
 path = "src/main.rs"
 
 [dependencies]
+oj_cache.workspace = true
 oj_compiler.workspace = true
 oj_env.workspace = true
 oj_config.workspace = true
--- a/crates/oj/src/build.rs
+++ b/crates/oj/src/build.rs
@@ -583,6 +583,7 @@ fn expand_css_via_sidecar(root: &Path, css_file: &Path) -> anyhow::Result<String
             css_file.to_str().unwrap(),
             root.to_str().unwrap(),
         ])
+        .env("NODE_COMPILE_CACHE", oj_server::node_compile_cache(root))
         .current_dir(root)
         .output()
         .context("node not found for tailwind/postcss build")?;
@@ -614,6 +615,7 @@ fn svelte_via_sidecar(root: &Path, file: &Path) -> anyhow::Result<String> {
     .to_string();
     let mut child = std::process::Command::new("node")
         .arg(&script)
+        .env("NODE_COMPILE_CACHE", oj_server::node_compile_cache(root))
         .current_dir(root)
         .stdin(std::process::Stdio::piped())
         .stdout(std::process::Stdio::piped())
@@ -658,6 +660,7 @@ fn preprocess_via_sidecar(root: &Path, css_file: &Path) -> anyhow::Result<String
     .to_string();
     let mut child = std::process::Command::new("node")
         .arg(&script)
+        .env("NODE_COMPILE_CACHE", oj_server::node_compile_cache(root))
         .current_dir(root)
         .stdin(std::process::Stdio::piped())
         .stdout(std::process::Stdio::piped())
@@ -1974,6 +1977,7 @@ pub(crate) async fn build_ssr_app(
         let out = std::process::Command::new("node")
             .arg(&script_path)
             .arg(serde_json::to_string(&paths)?)
+            .env("NODE_COMPILE_CACHE", oj_server::node_compile_cache(root))
             .current_dir(out_dir)
             .output()
             .context("node not found for prerender")?;
--- a/crates/oj/src/ssr_dev.rs
+++ b/crates/oj/src/ssr_dev.rs
@@ -86,17 +86,21 @@ async fn spawn_runner(root: &Path, base: &str, entry_abs: &Path) -> anyhow::Resu
     let script = dir.join("runner.mjs");
     std::fs::write(&script, oj_server::SSR_RUNNER_JS)?;
 
-    let mut child = tokio::process::Command::new("node")
-        .args([
-            "--experimental-vm-modules",
-            &script.to_string_lossy(),
-            base,
-            &entry_abs.to_string_lossy(),
-        ])
-        .current_dir(root)
-        .stdin(Stdio::piped())
-        .stdout(Stdio::piped())
-        .stderr(Stdio::inherit())
+    let mut cmd = tokio::process::Command::new("node");
+    cmd.args([
+        "--experimental-vm-modules",
+        &script.to_string_lossy(),
+        base,
+        &entry_abs.to_string_lossy(),
+    ])
+    .current_dir(root)
+    .stdin(Stdio::piped())
+    .stdout(Stdio::piped())
+    .stderr(Stdio::inherit());
+    if let Some(v8) = oj_server::node_compile_cache_opt_in(root) {
+        cmd.env("NODE_COMPILE_CACHE", v8);
+    }
+    let mut child = cmd
         .spawn()
         .map_err(|e| anyhow::anyhow!("could not spawn SSR runner (node): {e}"))?;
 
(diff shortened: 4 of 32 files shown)


raphamorim:
Landed on `main` as 88a55c6 (squashed from the 3 branch commits, authored by @aeriksson-lovable). The offset-based pack (`packGet` reads a line by [offset,len] then re-verifies `e.k === key` after parse, degrading to a per-entry miss on any read/parse failure) bounds runner memory to offsets not sources. Nice touch: `8e74c55` makes the runner tolerate a minimal stub loader — which supersedes the test-stub patch I'd added in #19, so I restored `runner-protocol.test.mjs` to the minimal stub, and it now genuinely exercises that tolerance (19/19 pass). `start.mjs` passes; `OJ_SSR_MEM_STATS` is off-by-default. Thanks André!

PR #22, opened 2026-08-21
host the SSR plugin container in the plugin host process

aeriksson-lovable:
Part 13 of 16 in a series. Depends on #21 — this PR's diff includes the commits of the PRs below it in the series; to see only this part: https://github.com/aeriksson-lovable/oj/compare/pr/10-memory-pass...pr/11-single-plugin-host. Series overview in #12.

---

Depends on #21 ("serve SSR load-pack hits by file offset to bound
runner memory"): all baseline numbers below are measured with that PR
applied.

## Problem

In start mode, `oj dev` spawns two long-lived Node.js processes:

- The **plugin host** (`crates/oj_server/src/assets/plugin-host.mjs`).
  It evaluates the app's Vite config and runs plugin hooks for the client
  environment.
- The **runner**. It executes the app's SSR server code. oj spawns it from
  `crates/oj/src/start_dev.rs`.

The runner registers a synchronous module loader
(`crates/oj_server/src/assets/start/loader.mjs`) through
`module.registerHooks()`. These hooks must return values, not promises.
When a module is not in the on-disk transform cache (a **miss**), the loader
must run Vite plugin hooks (`resolveId`, `load`, `transform`) for it.

Today the loader gets those hooks from its own copy of the plugin universe.
`crates/oj_server/src/assets/start/container-bridge.mjs` spawns a worker
thread (`crates/oj_server/src/assets/start/container-host.mjs`). The worker
calls `loadPluginContainer()` in
`crates/oj_server/src/assets/start/vite-plugin-bridge.mjs`. That function
evaluates the Vite config a second time and instantiates every plugin again,
with every transitive import (wrangler, miniflare, prettier, babel). The
bridge blocks on `Atomics.wait` over a `SharedArrayBuffer` until the worker
replies.

So one `oj dev` loads the plugin universe twice: once in the plugin host,
once in the runner's worker. On our test app (described under Performance),
the runner's duplicate copy costs 218 MB of JS heap plus 125 MB of external
memory — about 343 MB of RSS that holds no unique state.

The duplication cannot be removed by sharing the `SharedArrayBuffer`. A
`SharedArrayBuffer` is shared memory between threads of one process. It
cannot be transferred to another OS process, so `Atomics.wait` cannot block
on writes from the plugin host. A cross-process blocking primitive is needed
instead. A named pipe (FIFO) read is that primitive: `fs.readSync` on a FIFO
blocks in the kernel until the peer writes, and it keeps the loader hooks
synchronous.

## Change

One process now owns all plugin containers. The plugin host builds an
SSR-environment container from the plugin list it already evaluated, and the
runner's loader reaches it over a FIFO pair.

Boot sequence:

1. `start_dev()` in `crates/oj/src/start_dev.rs` calls
   `prepare_ssr_bridge()` (new, in `crates/oj_server/src/plugins.rs`)
   **before it spawns any child process**. `ssr_bridge_dir()` resolves the
   bridge directory: the `OJ_SSR_BRIDGE_DIR` environment variable when set,
   otherwise a per-project directory in the **OS temp dir**
   (`std::env::temp_dir()/oj-ssr-bridge-<hash>`, `<hash>` a BLAKE3 digest of
   the app root path). `prepare_ssr_bridge()` creates that directory with
   mode 0700, deletes stale `disabled` and `ready` sentinel files from a
   previous run, and creates `req.fifo` and `rep.fifo` with `mkfifo(2)`
   (mode 0600). Doing this before any spawn removes two races: the loader
   can never observe a stale sentinel, and oj never unlinks a FIFO end that
   the runner already holds open.

   The bridge directory must never live inside the app tree. Build tools
   and file watchers scan the app tree, and a FIFO is a trap for any
   scanner that opens what it finds: `open(2)` blocks until a writer
   appears, `read(2)` blocks until data arrives. The motivating failure was
   Tailwind's oxide scanner (`Scanner::discover_sources`): in an app that
   does not gitignore `.oj-cache/`, it descended into the old
   `.oj-cache/start/ssr-bridge/` location, blocked forever on the FIFOs,
   and every CSS compile ate its full 30 s timeout and silently served raw
   uncompiled CSS. The FIFOs also persisted after exit and poisoned later
   dev-server runs (any oj binary, or plain Vite) in the same checkout.
   Two healing measures ship with the move: `prepare_ssr_bridge()` deletes
   a stale legacy `<app>/.oj-cache/start/ssr-bridge/` directory left by
   older builds, and oj writes `.oj-cache/.gitignore` containing `*` when
   it creates the cache dir (`ensure_cache_gitignore()`), keeping every
   cache artifact out of `git status` and out of gitignore-respecting
   scanners.
2. `DevServer::new()` in `crates/oj_server/src/lib.rs` passes the bridge
   directory to the start-mode client plugin host in its config JSON
   (`ssrBridge.dir`). Only this one host receives it; only one process may
   hold the FIFO read end. The runner receives the same path as
   `OJ_SSR_BRIDGE_DIR` in its spawn environment; `container-bridge.mjs`
   reads only that variable and treats its absence as "no bridge" — the
   path is decided once, in Rust, and never re-derived in JS.
3. `plugin-host.mjs` opens both FIFOs with `"r+"` (`O_RDWR`). An `O_RDWR`
   open of a FIFO never blocks waiting for a peer, and the read end never
   sees EOF when the runner closes its end. The host reads request frames
   from `req.fifo` and writes reply frames to `rep.fifo`. A frame is a
   4-byte little-endian length prefix followed by that many bytes of JSON
   (`{id, method, args}` requests; `{id, value}` or `{id, error}` replies).
4. `plugin-host.mjs` builds a second plugin container for the `ssr`
   environment through `createPluginContainer()`, a new factory split out of
   `loadPluginContainer()` in `vite-plugin-bridge.mjs`. The factory takes an
   already-evaluated plugin list, so the config runs **once** for both
   environments. Environment and `apply` filtering differ per container: on
   our test app the client container keeps 53 active plugins, the SSR
   container keeps 34.
5. The host accepts frames from process start, but request dispatch parks on
   a promise until the SSR container's `buildStart()` completes. The host
   then writes the `ready` sentinel file and answers the queued frames. This
   preserves the "buildStart before any load" ordering the worker gave, and
   the `ready` file is the cross-process replacement for the worker's
   in-memory bootstrap flag (`bootstrapDone()` in the bridge facade checks
   it without blocking).
6. `container-bridge.mjs` keeps its synchronous facade API unchanged
   (`resolveId`, `load`, `transform`, `transformUserCode`, `env`, `heap`).
   Each call writes one frame to `req.fifo` and blocks in `fs.readSync` on
   `rep.fifo` until the matching reply arrives. The connection is **lazy**:
   the first plugin-served miss triggers `connect()`. A warm boot where
   every module hits the cache never opens a FIFO and never waits for the
   host. The `heap()` diagnostic also never forces a connection.
7. `connect()` probes host liveness with a non-blocking `O_WRONLY` open of
   `req.fifo`. That open succeeds only once the host holds the read end. The
   bridge polls every 25 ms up to 300 s (the same horizon the worker bridge
   used for a cold config eval plus `buildStart`), and checks the `disabled`
   sentinel on every iteration.
8. Every path that leaves the FIFOs unserved writes the `disabled` sentinel:
   a non-Unix platform (no `mkfifo`), a `mkfifo` failure, a plugin host that
   fails to start, a host shut down because native filtering left zero
   active plugins, and a host whose own FIFO open fails. When the bridge
   sees `disabled`, every facade call returns `null`. The loader already
   treats `null` as "no plugin claims this module" and serves the module
   through its own native pipeline. Without the sentinel, the loader's first
   miss would block for 300 s on a FIFO that no process will ever answer.
9. `container-host.mjs` is deleted. The runner spawns no plugin worker.

Files changed:

| Path | Change |
| --- | --- |
| `crates/oj/src/start_dev.rs` | Call `prepare_ssr_bridge()` before any child spawns; pass `OJ_SSR_BRIDGE_DIR` to the runner's environment; remove the bridge dir on SIGINT/SIGTERM. |
| `crates/oj_server/src/plugins.rs` | New bridge lifecycle: `ssr_bridge_dir()` (OS temp dir, `OJ_SSR_BRIDGE_DIR` override), `prepare_ssr_bridge()`, `ensure_ssr_bridge()`, `disable_ssr_bridge()`, `remove_legacy_ssr_bridge()`, `cleanup_ssr_bridge()`, `mkfifo` wrapper. |
| `crates/oj_server/src/lib.rs` | Pass `ssrBridge.dir` to the start-mode client host only; write `disabled` on host failure and on zero-plugin shutdown; `ensure_cache_gitignore()`; drop the `container-host.mjs` embedded asset. |
| `crates/oj_server/src/assets/plugin-host.mjs` | FIFO frame server; SSR container over the shared plugin list; dispatch parked until `buildStart` completes; `ready` sentinel; `process.env` delta capture for the loader. |
| `crates/oj_server/src/assets/start/vite-plugin-bridge.mjs` | Split `createPluginContainer()` (container over an evaluated plugin list) out of `loadPluginContainer()` (config eval + container). |
| `crates/oj_server/src/assets/start/container-bridge.mjs` | Replace the worker + `Atomics.wait` transport with the FIFO client: lazy `connect()`, framed synchronous calls, exact-length reads, `disabled`/`ready` sentinel handling. |
| `crates/oj_server/src/assets/start/container-host.mjs` | Deleted. |
| `crates/oj_server/Cargo.toml`, `Cargo.lock` | Add `libc` (Unix targets only) for `mkfifo`. |

## Correctness

- **Hook ordering.** The host answers no request before the SSR container's
  `buildStart()` completes. This matches the worker, which queued messages
  until bootstrap finished.
- **Environment variables.** Config evaluation mutates `process.env` (for
  example, `VITE_*` values derived from git state). The host snapshots
  `process.env` at process start and reports the delta through the `__env`
  method, exactly as the worker did. The loader's persisted env-delta cache
  is unchanged.
- **Failure fallback.** Every failure path converges on the `disabled`
  sentinel, and the bridge then returns `null` from every hook. That is the
  same observable behavior the worker had when its container failed to load
  (it also answered `null`). Modules are then served without plugin hooks;
  virtual modules get the existing stub with an explanatory error.
- **Host exit mid-run.** If `readSync` returns 0 bytes, the host has exited.
  The bridge marks itself down and throws
  `"oj: SSR plugin bridge closed (plugin host exited)"` for that call; later
  calls return `null`. `EAGAIN`/`EINTR` are retried.
- **Stale state across runs.** `prepare_ssr_bridge()` deletes old sentinels
  and re-creates both FIFOs on every boot, before any child spawns. It also
  removes a legacy in-tree `<app>/.oj-cache/start/ssr-bridge/` directory
  left by older builds, so a poisoned checkout heals on the next boot.
- **FIFO lifetime.** Unlinking a FIFO right after the first connection is
  not safe: the bridge connects lazily (a full-cache-hit boot never opens
  the FIFOs) and a runner reload reconnects by path, so the paths must
  outlive the first connection. The model is cleanup at graceful exit
  (SIGINT/SIGTERM removes the bridge dir) plus stale cleanup at the next
  boot; an ungraceful death leaves the pair in the OS temp dir, where no
  source scanner ever sees it.
- **Partial writes.** Both sides loop until the whole frame is written and
  read exact byte counts on receive. The host serializes replies through a
  write chain, so reply frames never interleave. There is one client (the
  runner), and its calls are synchronous, so requests never interleave
  either.
- **Non-Unix platforms.** `prepare_ssr_bridge()` writes `disabled` and
  returns `None` where FIFOs do not exist. The runner behaves as if the app
  had no serving plugins. This is the only platform difference.
- **Known limitation (intentional).** After a connection is established,
  a call has no per-request read deadline: `readSync` blocks until the host
  replies or exits. The previous worker bridge applied its 300 s deadline to
  every call. In this change only `connect()` carries the 300 s deadline. A
  plugin hook that never settles therefore blocks the runner instead of
  timing out. Host death is still detected immediately through EOF. We kept
  the transport minimal. A follow-up can add a read deadline with `poll(2)`
  semantics through a small native helper or a watchdog; we are open to
  that.

## Performance

`.oj-cache` directory is populated from a previous run. Cold means the cache
directory is empty. All measurements ran serially, never concurrently.

Memory is settled RSS after boot, warm full-hit run, on this app:

- Runner: 886 MB → 372 MB (−514 MB). The duplicate plugin universe is gone.
- Plugin host: +100 MB. The second container shares the already-imported
  plugin modules; the growth is container and transform state, not a second
  module universe.
- Total across all oj processes: 1,597 MB → 1,183 MB (−26 %).

Latency, same app:

- TTFC is unregressed, warm and cold.
- A cold (full-miss) boot pushes approximately 50,000 calls through the
  bridge. Each call completes in under 1 ms. Cold TTFC stays at 7 s.

## How to test

1. On macOS or Linux, run `oj dev` in a TanStack Start app whose Vite config
   registers plugins that serve modules (for example an MDX or SVG plugin).
2. Check stderr for `plugin host: ssr container ready (N plugin(s), shared
   config eval)`.
3. Run `ls "${TMPDIR:-/tmp}"/oj-ssr-bridge-*/`. It contains `req.fifo`,
   `rep.fifo`, and, after the container boots, `ready`. Confirm
   `.oj-cache/start/ssr-bridge/` does **not** exist under the app, and that
   `.oj-cache/.gitignore` exists and contains `*`.
4. Request an SSR page. Compare the HTML with the parent commit's output; it
   is identical.
5. Delete `.oj-cache`, boot again (cold), and confirm plugin-served modules
   (virtual modules, MDX, `.svg?react`) still resolve.
6. Fallback path: run an app whose plugins are all replaced natively. oj
   prints `served natively`, writes the `disabled` sentinel, and the runner
   boots and serves without touching the FIFOs.
7. Memory: compare per-process RSS (`ps -o rss,command | grep node`) after a
   warm boot on this commit and on the parent commit.


raphamorim:
Landed on `main` as 41da5ba (squashed from the 2 branch commits, authored by @aeriksson-lovable). Fork-reviewed the cross-process FIFO bridge in depth (plus `cargo check` + `node --check`): framing is desync-proof (`readExact` loops on partial reads, host advances past malformed frames), no deadlock (300s connect deadline + `disabled` sentinel; a host crash closes its write end → clean throw), requests are strictly serial with id-match, and every unserved path (spawn failure / zero plugins / non-Unix) writes `disabled` so the loader never blocks on a FIFO nobody answers. The self-SIGKILL is confirmed necessary — `process.exit` would `uv_thread_join` on the parked `req.fifo` read the host holds open `r+`. `oj_server` (44 tests incl. the new bridge tests), `start.mjs`, and `plugin-middleware` pass. Thanks André!

PR #23, opened 2026-08-21
serve recorded plugin container answers during the boot module walk

aeriksson-lovable:
Part 14 of 16 in a series. Depends on #22 — this PR's diff includes the commits of the PRs below it in the series; to see only this part: https://github.com/aeriksson-lovable/oj/compare/pr/11-single-plugin-host...pr/12-eager-preeval. Series overview in #12.

---

> Depends on #22 ("host the SSR plugin container in the plugin host
> process"). This PR speculates the container answers that #22's FIFO
> bridge carries, and revalidates them against its live container.

## Problem

The runner is the Node.js process that executes the app's server code in
start mode. oj spawns it from `crates/oj/src/start_dev.rs`. On boot, the
runner imports the server entry and walks its module graph.

The plugin host is a separate Node.js process
(`crates/oj_server/src/assets/plugin-host.mjs`). It evaluates the app's Vite
config once and runs plugin hooks. The plugin container is the component
inside the host that serves the SSR environment's `resolveId`, `load`, and
`transformUserCode` hooks. The runner's module loader
(`crates/oj_server/src/assets/start/loader.mjs`) calls the container
synchronously over a FIFO bridge
(`crates/oj_server/src/assets/start/container-bridge.mjs`): one
length-prefixed JSON frame per call, one framed reply.

The container must bootstrap before it can answer any call. Bootstrap means:
evaluate the config file, then run every plugin's `buildStart` hook. The host
writes a `ready` sentinel file when bootstrap completes; the bridge's
`bootstrapDone()` reads it.

(M-series), release build of oj. Warm means the `.oj-cache` directory is
populated from a previous run. In this feature's own measurement session
(warm boots, before composition with #22 — the container then ran in a
worker thread inside the runner):

- The runner starts its module walk at approximately +50 ms after spawn.
- The walk reaches its first container call and stops. It waits approximately
  1.2 s, because container bootstrap (config eval + `buildStart`) takes
  approximately 1.35 s.
- After bootstrap, all container calls together cost approximately 90 ms.

The cost is waiting, not working. The walk and the bootstrap run serially,
although they share no data until the first answer.

Composition with #22 adds a second cost: every container call now pays
a FIFO round trip between two processes. On this app, the module walk issues
thousands of container calls (see Performance). Both costs have the same fix:
answer from a recording, and check the recording afterwards.

Two loader paths also called the container on every boot without need:

- Every non-`node_modules` `.js`/`.mjs` file gets a `load` call to check
  whether a plugin claims it. On this app, almost none are claimed.
- Every user file goes through `transformUserCode`. The pipeline already
  treats this hook as pure in `(path, source)`: its output is persisted in
  the content-keyed load cache.

## Change

The loader records container answers and serves the recording during the
boot module walk. After the walk, a revalidation step checks every served
answer against the live container.

1. The loader wraps the container (`speculativeContainer` in `loader.mjs`).
   The wrapper intercepts `resolveId`, `load`, and `transformUserCode`.
2. Every answer is recorded in `.oj-cache/ssr-bridge-pack.jsonl`. The file's
   header line carries the cache epoch: a hash of the Node.js version, the
   lockfile, `package.json`, the Vite/oj config files, the loader's own
   sources, and the define/env inputs. An epoch mismatch discards the file.
3. Each record's key is a sha256 hash of the method name and its arguments.
   For a `load` whose id is a readable file, the key also includes the
   file's current on-disk bytes. An edited file therefore misses. For
   `transformUserCode`, the source text is an argument, so the key is fully
   content-determined.
4. A speculation window opens when the loader starts. While the window is
   open and the key is in the recording, the wrapper returns the recorded
   answer and notes the call in a `speculatedCalls` map. Otherwise it calls
   the live container and records the fresh answer.
5. The window does not close when bootstrap completes. It closes at the
   first revalidation. #22's shared host finishes bootstrap at
   approximately 1.5 s (warm boot, test app). A bootstrap-closed window
   therefore reopened the FIFO path in the middle of the walk and sent
   approximately 3,000–4,000 recordable calls over it. That variant measured
   2.9 s first-200 against 1.7 s (see Performance). The trust model does not
   change: the window only defers live calls, and every deferred answer is
   checked in step 6.
6. On boot, `start_dev.rs` sends the runner one synthetic `GET /` and
   discards the response (the existing prewarm render). When it completes,
   `start_dev.rs` sends a new `revalidate` command. The runner
   (`runner.mjs`) calls the loader's `revalidateSpeculation()`. It closes
   the window and replays every call in `speculatedCalls` against the
   now-live container. On any mismatch it updates the recording, re-imports
   the server entry under a bumped `?ojv=` version, and reports
   `reloaded: true`. `start_dev.rs` then pushes a live-reload to connected
   browser clients.
7. Bug found while testing, fixed here: `ojvirtual:` module URLs carried no
   version suffix. Node's module registry caches modules by URL. A reload's
   re-import therefore received the old virtual module instance — on any
   reload path, not only revalidation. `resolveUncached` now appends the
   same `?ojv=` suffix that file URLs already get, and `load` strips it.
8. The two no-need paths stop round-tripping on warm boots. The
   unclaimed-file check and the `transformUserCode` result are stored in the
   existing content-keyed persistent cache, under the same purity assumption
   the load cache already makes.

Files changed:

| Path | What changed |
| --- | --- |
| `crates/oj_server/src/assets/start/loader.mjs` | Recording file, key derivation, `speculativeContainer` wrapper, speculation window, `revalidateSpeculation()`, `?ojv=` suffix on `ojvirtual:` URLs, content-keyed unclaimed-file and `transformUserCode` caches. |
| `crates/oj_server/src/assets/start/runner.mjs` | Handles the `revalidate` command: replay, entry re-import, `reloaded` reply. Boot-phase timing lines under `OJ_BOOT_PHASES`. |
| `crates/oj/src/start_dev.rs` | Sends `revalidate` after the prewarm render; pushes a client live-reload when the runner reports `reloaded: true`. The reload channel now exists before the prewarm task. Boot-phase probe points. |
| `crates/oj_server/src/assets/start/container-bridge.mjs` | Per-call timing lines under `OJ_BOOT_PHASES` only. The bridge itself is #22's. |
| `crates/oj_server/src/assets/plugin-host.mjs` | One "bootstrap done" timing line under `OJ_BOOT_PHASES` only. |

## Correctness

The recording is never trusted permanently. Every answer served from it is
replayed against the live container after the prewarm render. The window
then closes; later calls reach the live container or the content-keyed
persistent caches.

For file-backed `load` answers and for `transformUserCode`, the key contains
the current content (on-disk bytes or the source argument). An edited file
misses the recording, and the call blocks on the live container. This is the
conservative path: no recorded answer is served for changed content.

Virtual ids have no file content to key on. They are the reason the
revalidation step exists. A mismatch during replay updates the recording,
re-imports the entry, and pushes a live-reload.

This leaves a bounded staleness window between the first response and the
completed revalidation. In this feature's own measurement session (warm
boot, test app), revalidation completed approximately 0.5 s after the server
started listening. On the final composed branch the gap is smaller: the
phase probe shows listening at 1.53 s and revalidation done at 1.62 s (warm
boot, test app). A request inside that window can render from a stale
virtual-module answer. The pushed live-reload then corrects every connected
client. The user-visible effect is the same class as an HMR full reload. We
verified this with live edits: stop the server, edit a source file that a
virtual module derives from (an i18n message file on the test app), start
`oj dev`, load the page. The runner reports `reloaded: true` and the browser
shows the fresh content. The same holds on revert.

Other cases:

- Missing, corrupt, or epoch-mismatched recording file: the recording starts
  empty. Every call blocks on the live container. Behavior equals the code
  before this change.
- `OJ_SSR_LOADER_CACHE=off`: no epoch, no recording, no speculation.
- Revalidation failure (runner write error, parse error): `start_dev.rs`
  treats it as "no reload". The recording is not marked valid.
- Keeping the window open past bootstrap means a call issued after bootstrap
  can also be answered from the recording, until revalidation. Such answers
  sit in `speculatedCalls` like all others and get the same replay.
- The `ojvirtual:` version fix also repairs the pre-existing watcher-reload
  path. Before it, any reload could serve stale virtual modules from Node's
  registry.

Intentionally not covered: bootstrap itself is never skipped. Config eval
and `buildStart` always run in the plugin host; plugin side effects still
happen. Only the module walk stops waiting for them. `transform` and `env`
calls are not speculated. #22's `bootstrapDone()` sentinel stays
available on the bridge, but this PR's gate deliberately does not use it
(see Change step 5).

## Performance

First-200 means the wall-clock time from `oj dev` start until the first
`GET /` returns HTTP 200. "Entry evaluated" is the runner's own timestamp
for finishing the server-entry import, relative to runner spawn. All runs
are warm boots of the test app described above, measured serially, never
concurrently.

This feature's own measurement session (before composition with #22;
container in an in-runner worker thread):

- Entry evaluated: +1,519 ms → +370 ms.
- First-200: 2.73 s → 1.70 s (6 runs, range 1.66–1.79 s).

Negative result, measured during composition: a window that closes at
bootstrap-done regressed the composed tree to 2.9 s first-200, against 1.7 s
with the revalidation-closed window. The shared host bootstraps at
approximately 1.5 s, and the early close sent approximately 3,000–4,000
recordable calls over the FIFO mid-walk. This is why step 5 keeps the window
open until revalidation.

Final composed branch (all PRs in this series), phase probe on one warm
boot: listening 1.53 s, prewarm done 1.58 s, revalidated
(`reloaded=false`) 1.62 s, first-200 at 1.68 s (1.76 s measured from process
start). The walk makes zero live bridge calls before the prewarm render.
The loader caches converge to 5,980/5,980 load hits and 44,157/44,157
resolve hits. Steady warm TTFC on the composed branch is 2 s.

Context for the design: after bootstrap, all container calls together cost
approximately 90 ms in the pre-composition session. Caching answers without
overlapping bootstrap would therefore save little there. The win comes from
not waiting — and, after #22, from not paying per-call FIFO round trips
during the walk.

## How to test

1. Run `oj dev` on an app with a Vite config. Stop it. Confirm
   `.oj-cache/ssr-bridge-pack.jsonl` exists.
2. Run `oj dev` again with `OJ_BOOT_PHASES=1`. The boot is faster. "entry
   evaluated" now lands before the container's "bootstrap done" line, and no
   `bridge:` call lines appear before "prewarm complete".
3. Staleness check: stop the server. Edit a file that a virtual module
   derives from (for example a locale JSON that an i18n plugin compiles).
   Start `oj dev` and open the page. The revalidation step detects the
   mismatch, the runner logs `revalidated (reloaded=true)` under
   `OJ_BOOT_PHASES=1`, and an open browser tab receives a live-reload with
   the fresh content.
4. Regular-file check: stop the server. Edit a plugin-claimed source file.
   Start `oj dev`. The first `load` for that file misses the recording and
   blocks on the live container. No stale content is served.
5. Delete `.oj-cache/ssr-bridge-pack.jsonl` or set `OJ_SSR_LOADER_CACHE=off`.
   Boot still works; every call blocks on the live container as before.


raphamorim:
Landed on `main` as dc1d6f7 (squashed from the 2 branch commits, authored by @aeriksson-lovable). Reviewed the speculation: file-`load` keys include on-disk bytes so edited files go live, and `revalidateSpeculation` replays every speculated call against the live container after the discarded prewarm — updating the pack and returning `stale` to bump the version and push a client reload — so staleness is bounded to the prewarm plus a small self-correcting window. `transformUserCode`/unclaimed-file checks join the content-keyed cache, and the `?ojv=` versioning fixes virtual-module reload refresh. oj (36) + oj_server (44) + runner-protocol + `start.mjs` pass. Thanks André!

PR #24, opened 2026-08-21
frame loader pack records with a length and hash prefix

aeriksson-lovable:
Part 15 of 16 in a series. Depends on #23 — this PR's diff includes the commits of the PRs below it in the series; to see only this part: https://github.com/aeriksson-lovable/oj/compare/pr/12-eager-preeval...pr/13-pack-integrity. Series overview in #12.

---

> Depends on #19 ("run SSR loader hooks in-thread via
> module.registerHooks" — it introduces the load pack) and #23
> ("serve recorded plugin container answers during the boot module walk"
> — it introduces the bridge pack).

## Problem

The loader (`crates/oj_server/src/assets/start/loader.mjs`) persists three
append-only JSONL caches, called packs:

- `ssr-load-pack.jsonl` — compiled module sources.
- `ssr-resolve.jsonl` — (parentURL, specifier) → URL resolutions.
- `ssr-bridge-pack.jsonl` — recorded plugin-container answers.

Each pack line is one JSON record. The packs are append-only, and the
loader flushes them from timers and from explicit flush points. A process
killed mid-write leaves a torn last line. Two flush paths could also race:
`revalidateSpeculation()` rewrites a whole pack with `writeFile` while a
timer-driven `appendFile` is in flight, and the kernel can interleave the
two writes mid-file. We reproduced this: one killed boot left 713 KB of
garbage in the middle of the bridge pack, not at its tail.

Before this change the loader parsed each line with a bare `JSON.parse`
and silently skipped failures. A torn or garbled pack therefore loaded
partially — one observed instance persisted 2,951 of 5,963 entries — and
the next boot ran half-warm with no signal about why.

## Change

Every pack record gains a self-describing frame, and every failure mode
gets one defined recovery action.

1. A record line is now `LLLLLLLL HHHHHHHHHHHHHHHH PAYLOAD`: an 8-hex
   payload byte length, a 16-hex prefix of the payload's sha256, and the
   JSON payload. The framing helpers live in
   `crates/oj_server/src/assets/start/loader-util.mjs`.
2. The header record (line one of each pack) gains a format version:
   `{ fmt: 2, epoch }`. `epoch` is the existing cache epoch hash.
3. On load, failures map to actions:
   - Framing failure (bad length field, line shorter than its length):
     truncate the file to the last good record. A torn tail is harmless
     by construction; everything before it is intact.
   - Hash or JSON-parse failure on one record: skip that record only.
   - Format-version mismatch: delete the pack and rebuild it.
   Every action logs one structured stderr line and degrades to a cache
   miss. The loader never crashes on a bad pack.
4. The load pack keeps the offset-index + pread design from #21:
   init frames the records and indexes (payload offset, length, hash);
   a record's hash is checked when its payload is actually read.
   `OJ_CACHE_VERIFY=full` verifies every record at init instead.
5. The resolve-pack and bridge-pack flushes now run on a serialized
   promise chain, the same discipline the load pack already used. This
   removes the append-vs-rewrite race described above.

Files changed:

| Path | What changed |
| --- | --- |
| `crates/oj_server/src/assets/start/loader-util.mjs` | Framing helpers: encode a record, scan a framed file, verify a payload hash. |
| `crates/oj_server/src/assets/start/loader.mjs` | All three packs write and read framed records; header carries `fmt`; per-failure recovery actions; serialized flush chains. |
| `e2e/unit/pack-format.test.mjs` | New. 11 tests: framing round-trip, torn tail, flipped byte in default and full modes, format-mismatch delete, torn header, epoch keep, resync skip, stderr contract. |

## Correctness

The per-entry cache stays authoritative for load results, and the live
container stays authoritative for bridge answers. A record the framing
rejects is not repaired; it is dropped, and the entry falls back to the
per-entry cache or a recompute. Nothing wrong is ever served from a
damaged pack.

Failure handling, per case:

- Torn tail (kill mid-append): one truncate line, the prefix loads, the
  next boot is clean.
- One flipped payload byte: one skip line for that record; every other
  record loads.
- Old-format pack (written before this change): the `fmt` field is
  missing or wrong, so the pack is deleted and rebuilt during one boot.
- Torn header: the pack is unreadable by definition; delete and rebuild.
- The 16-hex sha256 prefix is an integrity check against corruption, not
  an authentication mechanism. 64 bits of hash make an accidental
  collision on a same-length corrupted payload negligible.

## Performance

(M-series), release build of oj. Warm means the `.oj-cache` directory is
populated from a previous run. All measurements ran serially, never
concurrently, on an isolated clone of the app (a live dev server on the
shared checkout would rewrite the packs mid-measurement).

- Warm parity: TTFC 3.2/3.2/3.2/3.3 s before this change and
  3.2/3.2/3.2/3.3 s after; hit rates identical at load 5,980/5,980
  (4 uncacheable) and resolve 44,157/44,157. The framing cost is below
  the measurement resolution.
- Old-format packs: 3 stderr delete lines, one rebuild boot (6.6–7.0 s;
  the epoch also rotates in that scenario, so the boot recompiles
  regardless), then full hits.
- Torn tails cut into all three packs with `dd`: one truncate line each
  (66,099 / 140 / 64 bytes dropped), the same boot completes at 3.2 s
  with full hits, and the next boot is clean.
- All 11,909 load-pack records flipped by one byte, default mode: 5,913
  read-time drop lines (exactly the records actually read — verification
  is lazy), load still 5,980/5,980 through the per-entry cache, TTFC
  3.2 s. Under `OJ_CACHE_VERIFY=full`: all 11,909 detected at init, zero
  read-time drops, TTFC 3.2 s.
- kill -9 at the moment the first pack flush hits disk: the next boot
  prints zero integrity lines and runs at full hits.

## How to test

1. Run `oj dev` twice on a TanStack Start app so the packs exist under
   `.oj-cache/`.
2. Truncate the last line of `ssr-load-pack.jsonl` mid-record and boot.
   stderr prints one truncate line; the boot serves at full hits.
3. Flip one byte inside a record of `ssr-resolve.jsonl` and boot. stderr
   prints one skip line; only that resolution misses.
4. Overwrite a pack's header line with garbage and boot. The pack is
   deleted and rebuilt; the boot succeeds.
5. Run `OJ_CACHE_VERIFY=full oj dev` on damaged packs: all damage is
   reported at init instead of on first read.
6. Run the unit tests: `node --test e2e/unit/pack-format.test.mjs`.


raphamorim:
Landed on `main` as a9e96b5 (authored by @aeriksson-lovable). Reviewed the framing: each record is `<8hex len><16hex sha256><payload>\n`, `scanPack` validates the versioned header, self-heals a torn tail by truncating to the last good offset, skips hash-fail/rejected records, and deletes on a format mismatch — every failure logs once and degrades to a miss, never a crash. The load pack keeps its offset-index + lazy pread hash check (`OJ_CACHE_VERIFY=full` verifies at init), and all three packs' flushes are now promise-chained so an append can't interleave with a rewrite. Ships its own `pack-format.test.mjs` (torn tail, flipped bytes, format mismatch) — 29 unit tests + `start.mjs` pass. Thanks André!

PR #25, opened 2026-08-21
namespace the cache root by format version (.oj-cache/v1)

aeriksson-lovable:
Part 16 of 16 in a series. Depends on #24 — this PR's diff includes the commits of the PRs below it in the series; to see only this part: https://github.com/aeriksson-lovable/oj/compare/pr/13-pack-integrity...pr/14-cache-versioning. Series overview in #12.

---

> Depends on #24 ("frame loader pack records with a length and hash
> prefix"). This PR completes the same integrity effort: #24 covers
> the JavaScript-side packs; this PR versions the cache layout and
> verifies the two remaining Rust-side stores.

## Problem

Two related problems.

First, layout skew. Every oj binary reads and writes one mutable
`.oj-cache/` directory in the app. Two oj versions with different store
formats can therefore contaminate each other's state: an old binary can
misread a new store, and a new binary can trust files an old binary
wrote. The per-store format numbers added earlier in this series protect
each store's own entries, but not the layout as a whole, and not files
outside any store (packs, sidecar scripts, the V8 cache).

Second, unverified artifacts. Two stores still read stored bytes without
content verification:

- The config-extract store (`crates/oj_cache/src/config_extract.rs`)
  parses its entry JSON and serves it.
- The codegen store (`crates/oj_cache/src/start_codegen.rs`) copies
  stored artifacts into the app tree — `src/routeTree.gen.ts` among them.

Both write entries with temp-file-plus-rename, which prevents torn
writes, but a same-size flipped byte in a stored artifact parses fine and
is served. For the codegen store the corrupted bytes land in the app's
source tree.

### Commit 1: version the cache root

1. The cache root becomes `.oj-cache/v<N>/`. One constant defines `N`:
   `oj_cache::CACHE_ROOT_VERSION` (starting at 1), resolved through the
   new `oj_cache::cache_root()`. Every store, sidecar script, JSONL pack,
   and the V8 compile cache lives under the versioned root. Only
   `.oj-cache/.gitignore` stays at the true root.
2. Boot heals the pre-v1 layout. `prepare_cache_root` (the function that
   was `ensure_cache_gitignore`) removes the known pre-v1 top-level
   entries: store directories, sidecar scripts, packs, and the
   persistent cache's two-hex shard directories. Unknown names are left
   alone; oj never read them, so oj does not delete them.
3. Node children receive the versioned root as `OJ_CACHE_ROOT` in their
   environment at every spawn site. `loader.mjs` falls back to its own
   location (oj materializes it under `<cache root>/start`), and
   `vite-extract.mjs` and `plugin-host.mjs` write their config bundle
   beside themselves. No `.mjs` file hardcodes the unversioned layout.

### Commit 2: verify the two remaining stores

4. `crates/oj_cache/src/integrity.rs` gains a self-verifying whole-file
   artifact: the payload's blake3 hex on line one, the payload after it,
   written through the existing `atomic_write`. It serves stores whose
   entry file name is a lookup key rather than a content hash.
5. The config-extract store writes and reads its entries through it. The
   entries are small, so every read re-hashes, regardless of
   `OJ_CACHE_VERIFY`.
6. The codegen store moves to format 2. `persist` records a
   `.files.json` manifest of (size, blake3) per stored artifact.
   `restore` re-hashes every stored artifact against the manifest before
   it writes anything into the app tree. An entry without a readable
   manifest is corrupt by definition. The byte-identical-skip rule on
   in-tree writes stays.

Files changed:

| Path | What changed |
| --- | --- |
| `crates/oj_cache/src/lib.rs` | `CACHE_ROOT_VERSION`, `cache_root()`, `prepare_cache_root` heal of the pre-v1 layout. |
| `crates/oj/src/start_dev.rs`, `crates/oj/src/build.rs`, `crates/oj/src/ssr_dev.rs` | Spawn sites pass `OJ_CACHE_ROOT`; paths go through `cache_root()`. |
| `crates/oj_server/src/lib.rs`, `optimize.rs`, `plugins.rs`, `sidecar.rs` | Same for the server-side spawn sites. |
| `crates/oj_server/src/assets/{plugin-host,vite-extract}.mjs`, `assets/start/{loader,runner,css-host}.mjs` | Read `OJ_CACHE_ROOT`; no hardcoded unversioned paths. |
| `crates/oj_cache/src/integrity.rs` | Self-verifying whole-file artifact (blake3 header line). |
| `crates/oj_cache/src/config_extract.rs` | Entries stored and read through the self-verifying artifact. |
| `crates/oj_cache/src/start_codegen.rs` | Format 2: `.files.json` manifest; re-hash before any in-tree restore; corruption-injection tests. |
| `crates/oj_cache/src/start_bundle.rs` | Store path through `cache_root()`. |
| `e2e/dep-optimize.mjs`, `e2e/ssr-dev.mjs` | Test paths follow the versioned root. |

## Correctness

- One constant controls the layout version. Bumping
  `CACHE_ROOT_VERSION` abandons the whole old layout at once; no store
  can be read by a binary that expects a different layout.
- The heal deletes only names oj itself created in the pre-v1 layout.
  A user file that happens to live in `.oj-cache/` survives.
- Every verification failure logs one structured line, deletes the
  entry, and degrades to a miss. Corruption-injection unit tests pin
  flip → miss + recompute, never wrong output, for both stores.
- The first boot after upgrading is a cold boot: the versioned root
  starts empty. This is the intended cost of the migration, paid once.
- Node children that receive no `OJ_CACHE_ROOT` (a script run by hand)
  fall back to a path derived from their own location, which oj places
  under the versioned root.

## Performance

concurrently, on an isolated clone of the app.

- TTFC ladder with a fresh cache: cold 9 s, warm 4 / 3 / 3 s. Steady
  warm hit rates: load 5,980/5,980 (4 uncacheable), resolve
  44,157/44,157 — parity with the parent commit.
- Legacy-layout heal: booting over a real pre-v1 cache (207 MB of
  shards, packs, stores, and scripts) leaves only `.gitignore` and
  `v1/`. That first boot is a cold boot (11 s including the heal); warm
  boots then converge as above.
- Corruption drills: a torn tail in the final load-pack record heals
  with one truncate line (2,594 B dropped) and a 3 s boot; one flipped
  resolve record skips with one line (resolve 44,156/44,157, 3 s boot);
  the next `OJ_CACHE_VERIFY=full` boot runs at full hits. Nothing wrong
  is ever served.
- Unit tests: `cargo test -p oj_cache -p oj -p oj_server` passes 102
  (55 + 23 + 24; 7 new for heal/versioning, self-verified framing, and
  per-store flip injection). `node --test e2e/unit`: 77/77.

## How to test

1. Run `oj dev` on an app with a populated pre-upgrade `.oj-cache/`.
   After boot, `ls .oj-cache/` shows only `.gitignore` and `v1/`.
2. Run `oj dev` again. The boot is warm; the caches hit.
3. Flip one byte in a stored codegen artifact under
   `.oj-cache/v1/start-codegen/route-tree/<key>/` and boot. The store
   logs the verification failure, deletes the entry, and regenerates.
   `src/routeTree.gen.ts` never receives corrupted bytes.
4. Corrupt a config-extract entry under `.oj-cache/v1/config-extract/`
   the same way; the next boot re-runs the extract.
5. Run the unit tests: `cargo test -p oj_cache`.


raphamorim:
Landed on `main` as f8ce9fd (squashed from the 2 branch commits, authored by @aeriksson-lovable). The versioned root (`.oj-cache/v1`, single `CACHE_ROOT_VERSION` constant) keeps binaries with incompatible store formats from contaminating each other; `prepare_cache_root` heals the known pre-v1 layout (leaving unknown names + `.gitignore`), and every spawn site passes `OJ_CACHE_ROOT`. The integrity refactor closes a real gap — config-extract/codegen were hand-rolled tmp+rename with no verification, so a same-size flip served wrong bytes (into the app tree for codegen); now `write/read_self_verified` (blake3 header, always re-hash) and codegen's `.files.json` manifest make any flip a miss + recompute. Full suite green: oj_cache (55) + oj (36) + oj_server (44), and `start`/`ssr-dev`/`dep-optimize` e2e. Thanks André — this completes the 16-PR series!

PR #26, opened 2026-08-21
route connect error middlewares per their arity

aeriksson-lovable:
Part 4 of 16 in a series. Depends on #14 — this PR's diff includes the commits of the PRs below it in the series; to see only this part: https://github.com/aeriksson-lovable/oj/compare/pr/03-serve-all-client-chunks...pr/03b-connect-error-middleware. Series overview in #12.

---

## Problem

`configureServer` plugins register middleware through
`server.middlewares.use(fn)`. In Vite, `server.middlewares` is a connect
instance. Connect defines two middleware kinds by function arity:

- A normal middleware takes three parameters: `(req, res, next)`.
- An error middleware takes four: `(err, req, res, next)`. Connect detects
  it with `fn.length >= 4`.

Connect dispatches the two kinds on separate paths. Normal dispatch calls
only normal middlewares. When a middleware passes an error to `next(err)`,
or throws synchronously, dispatch switches to the error path: it skips
normal middlewares and calls only error middlewares, as
`(err, req, res, next)`.

The middleware chain in `crates/oj_server/src/assets/plugin-host.mjs`
(`setupConfigureServer`) ignored arity. It called every registered
middleware as `(req, res, next)`. An error middleware then received
shifted arguments: `err` = the request, `req` = the response, `res` = the
`next` function, `next` = `undefined`. The first time it called
`next(err)` — the standard pass-through for an error it does not handle —
it threw `TypeError: next is not a function`. The chain caught the throw
and answered 500.

This breaks real configurations. We hit it with a dev-server plugin that
registers an error-handling middleware to convert a specific upstream
failure into a retry page. With that plugin active, every HTML request
through oj returned 500. The same config serves normally under Vite,
because connect never calls the error middleware on the normal path.

## Change

The dispatch loop in `setupConfigureServer` now follows the connect
contract:

1. `next(err)` carries the in-flight error. `next()` with no argument
   means no error is in flight.
2. The loop walks the stack from the current index. For each entry it
   first applies the existing path filter, unchanged.
3. It then compares the entry's kind against the error state:
   `fn.length >= 4` marks an error middleware. While no error is in
   flight, error middlewares are skipped. While an error is in flight,
   normal middlewares are skipped.
4. A matching normal middleware is called as `(req, res, next)`. A
   matching error middleware is called as `(err, req, res, next)`.
5. An error middleware that calls `next()` with no argument clears the
   error; dispatch resumes with normal middlewares at the next stack
   index. Calling `next(err2)` keeps the error path with the new error.
6. A synchronous throw in either kind becomes the in-flight error, as
   before.
7. End-of-stack behavior is unchanged. With an error in flight the
   response is 500 with the error stack, exactly what the old code
   answered on any `next(err)`. Without an error it stays the 404 with
   the `x-oj-fallthrough` header.

Files changed:

| Path | What changed |
| --- | --- |
| `crates/oj_server/src/assets/plugin-host.mjs` | The `next` closure in `setupConfigureServer` checks `fn.length` per stack entry and routes normal and error middlewares per the connect contract. |

## Correctness

- The skip rules are exactly connect's (`connect@3`, `proto.handle`): it
  selects on `fn.length >= 4` versus `error` being set, and calls
  `fn(err, req, res, next)` or `fn(req, res, next)` accordingly.
- Chains without error middlewares behave as before. A `next(err)` walks
  the rest of the stack, matches nothing (no entry has arity >= 4), and
  answers the same 500 with the error stack that the old code produced
  immediately.
- Chains with error middlewares gain the connect behavior: the error
  middleware only runs when an earlier middleware failed, and it receives
  the error as its first argument.
- `fn.length` counts declared parameters before the first default or rest
  parameter. A handler written as `(...args)` or with fewer named
  parameters reports a length below 4 and is treated as a normal
  middleware. This matches connect itself, which has the same limitation;
  middleware authors already must declare four named parameters for
  connect to treat a function as an error handler.
- Path-filtered entries (`use(path, fn)`) keep their prefix check on both
  dispatch paths.

## Performance

None expected: the change adds one arity comparison per stack entry per
request. Measured on a large real-world app: a TanStack Start application
with approximately 18,600 client modules, approximately 2,500 SSR modules,
and a 788-line Vite config. Hardware: macOS arm64 (M-series), release
build of oj. TTFC means the wall-clock time from `oj dev` start until
`GET /` returns HTTP 200 with HTML. All measurements ran serially.

- Plain config (53 plugins, no error middleware registered): warm TTFC
  3 s then 2 s, SSR loader caches at full hits (load 5,980/5,980, resolve
  44,157/44,157) — parity with the same binary before this change.
- Remote-development config of the same app (75 plugins, one error
  middleware): before this change every HTML request answered 500; after
  it, `GET /` answers 200 text/html. Cold TTFC 10.0 s; warm 3.7 s and
  2.6 s. Full 3,950-file / 83.2 MB transitive asset closure fetched with
  200s in 4.7 s (warm, fetched at concurrency 8). Steady RSS
  ~1.5 GB across the process group.

## How to test

1. Create a Vite app whose config registers both middleware kinds:

   ```js
   // vite.config.mjs (plugin)
   configureServer(server) {
     server.middlewares.use((req, res, next) => {
       if (req.url === "/boom") return next(new Error("boom"));
       next();
     });
     server.middlewares.use((err, req, res, next) => {
       res.statusCode = 503;
       res.end("handled: " + err.message);
     });
   }
   ```

2. `oj dev .` and request any normal page: it must answer 200. Before
   this change it answers 500 `TypeError: next is not a function` as soon
   as the request reaches the middleware chain.
3. Request `/boom`: the error middleware must answer it (503
   `handled: boom`).
4. Request a path whose error no middleware handles (the error middleware
   calls `next(err)`): the response must be 500 with the error stack (the
   unchanged end-of-chain behavior).

We ran these steps against this branch: `GET /` answers 200, `/boom`
answers 503 `handled: boom`, and an unhandled error answers the stack.

### DIFF
--- a/crates/oj/src/start_dev.rs
+++ b/crates/oj/src/start_dev.rs
@@ -25,7 +25,10 @@ struct StartState {
     proxy_prefixes: Vec<String>,
     plugin_mw_port: Option<u16>,
     runner: Arc<tokio::sync::Mutex<Runner>>,
-    client_bundle: PathBuf,
+    // The chunk set of the client build this server serves: swapped whole on
+    // watch rebuilds, so every /@oj-start/<chunk> resolves against exactly
+    // one build and a name outside it is a deliberate 404.
+    bundle: std::sync::RwLock<Arc<PinnedBundle>>,
     live_reload: PathBuf,
     workspace_root: PathBuf,
     reload_tx: broadcast::Sender<()>,
@@ -60,7 +63,12 @@ pub async fn start_dev(
     route_tree.await??;
     let bundle = {
         let (root, cache) = (root.clone(), cache.clone());
-        tokio::task::spawn_blocking(move || bundle_client_entry(&root, &cache))
+        tokio::task::spawn_blocking(move || -> anyhow::Result<PinnedBundle> {
+            bundle_client_entry(&root, &cache)?;
+            PinnedBundle::from_build_dir(&cache).ok_or_else(|| {
+                anyhow::anyhow!("client chunk index missing after successful bundle")
+            })
+        })
     };
     let built_fut = oj_server::DevServer {
         root: root.clone(),
@@ -71,7 +79,7 @@ pub async fn start_dev(
     }
     .build_app();
     let (bundle_res, built_res) = tokio::join!(bundle, built_fut);
-    bundle_res??;
+    let pinned = bundle_res??;
     resolver.await??;
     let built = built_res?;
     let runner = spawn_start_runner(&root, &cache).await?;
@@ -88,7 +96,7 @@ pub async fn start_dev(
         proxy_prefixes: built.proxy_prefixes.clone(),
         plugin_mw_port: built.plugin_mw_port,
         runner: Arc::new(tokio::sync::Mutex::new(runner)),
-        client_bundle: cache.join("client-entry.js"),
+        bundle: std::sync::RwLock::new(Arc::new(pinned)),
         live_reload: cache.join("live-reload.js"),
         workspace_root: workspace_root(&root),
         reload_tx: reload_tx.clone(),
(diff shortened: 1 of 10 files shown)


raphamorim:
Landed on `main` as 4b61ed2 (squashed, authored by @aeriksson-lovable). The arity gate `(fn.length >= 4) !== (err != null)` matches connect's contract exactly: normal handlers run only with no error in flight, error handlers as `(err,req,res,next)` only with one, and a no-arg `next()` clears the error. Verified `e2e/plugin-middleware.mjs` passes. Thanks André!

PR #27, opened 2026-08-21
stat stdin with bigint to avoid corrupting the realpath cache

aeriksson-lovable:
Part 5 of 16 in a series. Depends on #26 — this PR's diff includes the commits of the PRs below it in the series; to see only this part: https://github.com/aeriksson-lovable/oj/compare/pr/03b-connect-error-middleware...pr/03c-nested-require-resolution. Series overview in #12.

---

## Problem

Node resolves a bare specifier (`require("immer")`, `import "immer"`)
against the `node_modules` directories above the importer's file, after
resolving symlinks in the importer's own path. pnpm depends on this: every
package lives once under `node_modules/.pnpm/<pkg>@<version>/node_modules/
<pkg>`, and its dependencies are sibling symlinks in that same directory.
When an app pins one major version of a package and a dependency of a
dependency needs a different major version, both copies exist, and the
symlink resolution step is what routes each importer to its own copy.

Part 1 of this series (#12) breaks that step. It gave every long-lived
Node service oj spawns — the SSR runner, the plugin host, the CSS host,
the CSS/Tailwind/Svelte sidecars — one check at startup:
`fstatSync(0).isFIFO()`, to arm the orphan-exit handler only when stdin
is a pipe. `fstatSync` without options writes its result into a
process-wide stat buffer (`statValues` in `node:fs`) that
`fs.realpathSync` also reads. `realpathSync` walks a path one directory at
a time; for a directory it already has in its cache it skips the `lstat`
and instead consults that shared buffer to decide whether the walk hit a
pipe or a socket — and if the buffer says so, it stops and returns the
rest of the path unresolved (`node:fs`, `realpathSync`, the
`isFileType(statValues, S_IFIFO)` check inside the
`knownHard.has(base) || cache?.get(base) === base` branch). The buffer is
stale at that point: it still holds the FIFO entry from the startup check,
because the CJS loader stats files through `internalModuleStat`, which
never touches it. The first `require()` after boot that needs a realpath
therefore returns its input path unresolved — and `realpathSync` caches
that identity mapping, so every later resolution of the same path stays
wrong for the lifetime of the process.

The failure this produces: an app whose dependencies need a different
major version of a package than the app itself pins. Measured case: the
app pins `immer@10`; its chart library depends on `@reduxjs/toolkit`,
which needs `immer@11`. The chart library's CJS code calls
`require("@reduxjs/toolkit")`. The resolver finds the package through the
library's sibling symlink
(`node_modules/.pnpm/recharts@…/node_modules/@reduxjs/toolkit`), and —
with the realpath step disabled by the stale buffer — keeps the symlink
path instead of the real one. The package's entry is an ES module, so
`require()` loads it as an ESM graph rooted at the symlink URL. Its
`import { setUseStrictIteration } from "immer"` then walks `node_modules`
up from the symlink location, where pnpm placed the app's `immer@10`, not
the toolkit's own `immer@11`. Every SSR render fails with:

```
SyntaxError: The requested module 'immer' does not provide an export
named 'setUseStrictIteration'
```

The window opens exactly at the startup check. The SSR runner imports the
app's server entry before it stats stdin, so boot-time resolution is
correct; routes load lazily per request, after the stat, so the first
request poisons the cache and every request after it answers 500.

## Change

Request BigInt stats for the stdin check: `fstatSync(0, { bigint: true })`.
BigInt stat calls fill a separate buffer (`bigintStats`) and leave the
shared one untouched, so `realpathSync` never sees the FIFO entry.
`BigIntStats` has the same `isFIFO()` method; the orphan-exit behavior is
unchanged. One-token change at each of the seven startup checks:

| Path | What changed |
| --- | --- |
| `crates/oj_server/src/assets/start/runner.mjs` | stdin FIFO check → bigint |
| `crates/oj_server/src/assets/start/css-host.mjs` | same |
| `crates/oj_server/src/assets/plugin-host.mjs` | same |
| `crates/oj_server/src/assets/ssr-runner.mjs` | same |
| `crates/oj_server/src/assets/css-preprocess.mjs` | same |
| `crates/oj_server/src/assets/tailwind-sidecar.mjs` | same |
| `crates/oj_server/src/assets/svelte-compile.mjs` | same |

The underlying stale-buffer read is a Node bug (`realpathSync` trusts
`statValues` in a branch that performed no stat call); it is worth an
upstream report. This change removes the poisoning source that #12's
startup checks introduced in oj's own processes.

## Correctness

- `fs.fstatSync(fd, { bigint: true })` returns a `BigIntStats` object;
  `isFIFO()` reads the mode bits the same way. The check's result is
  identical for every fd kind.
- The change is in the startup path only. No per-request code changes.
- The fix removes the poisoning source inside oj's own services. App code
  that itself stats a pipe or a socket on the main thread can still
  trigger the Node bug; that is out of scope here and needs the upstream
  fix.
- All seven services get the same change, so no oj-spawned process leaves
  a FIFO entry in the shared buffer.

Standalone demonstration of the mechanism, no oj involved (macOS arm64,
Node 24.14.1):

```
mkdir -p /tmp/rp/real && cd /tmp/rp
echo "module.exports = 1;" > real/target.js
ln -s real/target.js link.js
echo | node -e '
  const fs = require("fs");
  require.resolve("./real/target.js");  // seeds the realpath cache
  fs.fstatSync(0);                      // stdin is a pipe: FIFO enters the shared buffer
  console.log(require.resolve("./link.js"));'
```

This prints `/tmp/rp/link.js` — the symlink, unresolved. Drop the
`fstatSync` line, or change it to `fstatSync(0, { bigint: true })`, and it
prints `/tmp/rp/real/target.js`.

## Performance

None expected: the change swaps which stat buffer one startup call fills.
Measured on a large real-world app: a TanStack Start application with
a 788-line Vite config that registers 53 plugins (75 in its
remote-development profile, the one that loads the chart library on the
SSR path). Hardware: macOS arm64 (M-series), release build of oj. TTFC
means the wall-clock time from `oj dev` start until `GET /` returns HTTP
200 with HTML. Warm means the `.oj-cache` directory is populated from a
previous run. All measurements ran serially, never concurrently.

- Before this change, with the remote-development profile: every request
  answers 500 with the immer SyntaxError above — at this commit and at
  every later optimization commit before the in-thread-hooks change
  (#19), which masks the bug by accident (in-thread loading keeps
  refreshing the shared buffer between stats).
- After this change, same profile, this commit only (no persistent caches
  exist yet at this commit): every request answers 200. TTFC 13.96 s
  cold, 14.33 s and 14.41 s on two repeat runs; full 3,950-file / 83.2 MB
  transitive asset closure fetched with 200s in 15.5 s; steady RSS
  ~2.3 GB across the process group.
- After this change, with the later optimization commits (#15, #16, #17,
  #18) rebased on top: TTFC 10.81 s cold, 5.32 s and 5.71 s warm; closure
  6.8–7.2 s warm; steady RSS ~2.4 GB. Identical closure size — same
  rendered app.
- The in-thread-hooks build (#19) re-booted with no code change: 200s,
  same closure, warm TTFC 4.89 s — unchanged behavior class. Its commits
  also apply cleanly on top of this change.

## How to test

1. Run the standalone snippet above on Node with `require(esm)` support:
   it must print the real path with this change's `bigint` form and the
   symlink path with the plain form.
2. pnpm app check: create a pnpm app that pins `immer@10` and depends on
   `recharts@3` (which depends on `@reduxjs/toolkit`, which needs
   `immer@11`). Add a page that renders any chart during SSR. `oj dev .`
   and request the page twice. Before this change the second request (and
   every one after) answers 500 with the `setUseStrictIteration`
   SyntaxError; after it, every request answers 200.
3. Orphan-exit check still works: `oj dev .`, `kill -9` the oj process,
   and confirm the runner/css-host/plugin-host processes exit instead of
   idling.

### DIFF
--- a/crates/oj/src/start_dev.rs
+++ b/crates/oj/src/start_dev.rs
@@ -25,7 +25,10 @@ struct StartState {
     proxy_prefixes: Vec<String>,
     plugin_mw_port: Option<u16>,
     runner: Arc<tokio::sync::Mutex<Runner>>,
-    client_bundle: PathBuf,
+    // The chunk set of the client build this server serves: swapped whole on
+    // watch rebuilds, so every /@oj-start/<chunk> resolves against exactly
+    // one build and a name outside it is a deliberate 404.
+    bundle: std::sync::RwLock<Arc<PinnedBundle>>,
     live_reload: PathBuf,
     workspace_root: PathBuf,
     reload_tx: broadcast::Sender<()>,
@@ -60,7 +63,12 @@ pub async fn start_dev(
     route_tree.await??;
     let bundle = {
         let (root, cache) = (root.clone(), cache.clone());
-        tokio::task::spawn_blocking(move || bundle_client_entry(&root, &cache))
+        tokio::task::spawn_blocking(move || -> anyhow::Result<PinnedBundle> {
+            bundle_client_entry(&root, &cache)?;
+            PinnedBundle::from_build_dir(&cache).ok_or_else(|| {
+                anyhow::anyhow!("client chunk index missing after successful bundle")
+            })
+        })
     };
     let built_fut = oj_server::DevServer {
         root: root.clone(),
@@ -71,7 +79,7 @@ pub async fn start_dev(
     }
     .build_app();
     let (bundle_res, built_res) = tokio::join!(bundle, built_fut);
-    bundle_res??;
+    let pinned = bundle_res??;
     resolver.await??;
     let built = built_res?;
     let runner = spawn_start_runner(&root, &cache).await?;
@@ -88,7 +96,7 @@ pub async fn start_dev(
         proxy_prefixes: built.proxy_prefixes.clone(),
         plugin_mw_port: built.plugin_mw_port,
         runner: Arc::new(tokio::sync::Mutex::new(runner)),
-        client_bundle: cache.join("client-entry.js"),
+        bundle: std::sync::RwLock::new(Arc::new(pinned)),
         live_reload: cache.join("live-reload.js"),
         workspace_root: workspace_root(&root),
         reload_tx: reload_tx.clone(),
@@ -253,9 +261,15 @@ fn spawn_start_watcher(root: PathBuf, cache: PathBuf, state: Arc<StartState>) {
             rt.block_on(async {
                 let (r, c) = (root.clone(), cache.clone());
                 let client = tokio::task::spawn_blocking(move || {
-                    let _ = bundle_client_entry(&r, &c);
+                    if bundle_client_entry(&r, &c).is_err() {
+                        return None;
+                    }
+                    PinnedBundle::from_build_dir(&c)
                 });
-                let (_, _) = tokio::join!(client, reload_runner(&state));
+                let (pinned, _) = tokio::join!(client, reload_runner(&state));
+                if let Ok(Some(pinned)) = pinned {
+                    *state.bundle.write().unwrap() = Arc::new(pinned);
+                }
             });
             let _ = state.reload_tx.send(());
             let modules = client_module_count(&cache);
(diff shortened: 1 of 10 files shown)


raphamorim:
Landed on `main` as d12aed4 (squashed, authored by @aeriksson-lovable). Confirmed the mechanism: `fstatSync(0, { bigint: true })` writes the separate BigInt stat buffer instead of the shared one `fs.realpathSync` consults on cache-hits, so the leftover FIFO entry can no longer poison nested (pnpm) symlink resolution; `BigIntStats.isFIFO()` keeps the orphan-exit behavior identical. `ssr-dev` + `preprocessors` pass. Thanks André!

PR #33, opened 2026-08-24
fix(build): fail on unresolved imports

williamhogman:
## Summary

- Treat unresolved Rollup imports as build failures instead of publishing a broken JavaScript bundle.
- Continue allowing dependencies explicitly declared external.
- Add a synthetic regression and execute it in the existing build job without adding a runner.

## Verification

- Confirmed the regression fails against main before the fix.
- cargo build --offline -p oj
- node e2e/build-unresolved.mjs

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -77,6 +77,8 @@ jobs:
           node-version: 22
       - run: npm ci --prefix playground
       - run: cargo run -p oj -- build playground
+      - name: reject unresolved build imports
+        run: node e2e/build-unresolved.mjs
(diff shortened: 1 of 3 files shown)


raphamorim:
Thanks @williamhogman — this landed on main via #134 with both of your commits preserved (ff0c8ac test, d0bef0b fix), plus one follow-up (f0baca2) so the failure matches Vite's observable behavior: the partial dist/ is dropped so a failed build never leaves a broken bundle behind, and the error names the build.rollupOptions.external escape hatch like Vite's does. Closing in favor of the landed version.

PR #34, opened 2026-08-24
fix(build): emit every configured HTML entry

williamhogman:
## Summary

- Honor every HTML entry declared in rollupOptions.input instead of only building index.html.
- Rewrite entry bundles, environment values, CSS, preload links, and HTML plugin transforms for each generated page.
- Keep generated pages inside the project output tree and cover entry ordering and containment with unit tests.
- Run the synthetic multi-page regression in the existing build job without adding a runner.

## Verification

- Confirmed the regression fails against main because the second page is missing.
- cargo build --offline -p oj
- node e2e/multi-page-build.mjs

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -77,6 +77,8 @@ jobs:
           node-version: 22
       - run: npm ci --prefix playground
       - run: cargo run -p oj -- build playground
+      - name: build every configured HTML entry
+        run: node e2e/multi-page-build.mjs
       - name: check outputs
         run: |
           test -f playground/dist/index.html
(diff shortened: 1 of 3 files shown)


raphamorim:
Thanks @williamhogman — the goal here is right, and main now has this end to end via the multi-HTML-input work that landed in the meantime: build.rollupOptions.input entries are all built (build.rs ~1476, falling back to index.html only when no input is configured), and every page goes through the same per-page pipeline this PR adds — env replacement, its own entry-chunk rewriting with modulepreload and split CSS links, transformIndexHtml per page, and a contained output path (an input outside root lands under its bare filename in dist/, never escaping the output tree). e2e/multi-html-input.mjs covers two nested pages with page-relative scripts. Since main already provides everything this PR sets out to do and the two implementations conflict over the same restructured region of build.rs, closing as superseded. If you hit a specific per-page behavior that main still gets wrong, a targeted PR against the current structure would be very welcome.

PR #35, opened 2026-08-24
fix(build): pass command and mode to Vite configuration

williamhogman:
## Summary
- Evaluate Vite configuration with command=build and the selected build mode instead of serve/development.
- Separate cached configuration extraction by command and mode.
- Preserve the existing development-server configuration behavior.
- Add a synthetic regression to the existing CI job without adding runners.

## Verification
- Confirmed the regression fails against main because production builds receive the development base.
- Confirmed the identical regression passes after the fix.
- node e2e/vite-config-command.mjs

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -46,6 +46,9 @@ jobs:
       - run: npx --prefix e2e playwright install --with-deps chromium
       - name: run e2e suite
         run: node e2e/run.mjs ${{ matrix.mode == 'bundle' && '--bundle' || '' }}
+      - name: Vite build command and mode
+        if: matrix.mode == 'unbundled'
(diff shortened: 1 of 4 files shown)


raphamorim:
Thanks @williamhogman — the diagnosis here was right, and the same fix landed independently on main in deca86c ("build: evaluate vite.config as build, adopt build block"): the extractor now receives command + mode, the build passes "build"/mode at the same three call sites, and the config-extract cache is keyed by both. That commit also adopts the normalized build block (outDir, sourcemap, minify, cssCodeSplit, target, ssr) and ships e2e/vite-config-build.mjs covering the command/mode branching. Since main already has everything this PR adds and merging would conflict on the duplicate signature change, closing as superseded.

PR #36, opened 2026-08-24
fix(config): reject Vite configurations that fail to load

williamhogman:
## Summary
- Fail builds and development-server startup when an existing Vite configuration cannot be loaded.
- Stop treating unsuccessful configuration extraction as a valid empty configuration.
- Preserve projects without a Vite configuration and explicit plugin files.
- Add a synthetic missing-plugin regression to the existing CI job without adding runners.

## Verification
- Confirmed the regression fails against main because an invalid Vite configuration is silently ignored.
- Confirmed the identical regression passes after the fix.
- node e2e/config-load-failure.mjs

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -46,6 +46,9 @@ jobs:
       - run: npx --prefix e2e playwright install --with-deps chromium
       - name: run e2e suite
         run: node e2e/run.mjs ${{ matrix.mode == 'bundle' && '--bundle' || '' }}
+      - name: Vite config load failures
+        if: matrix.mode == 'unbundled'
+        run: node e2e/config-load-failure.mjs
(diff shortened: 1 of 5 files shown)


raphamorim:
Thanks @williamhogman — the diagnosis was exactly right: a vite.config that failed to evaluate was being parsed into an empty configuration and the build carried on with defaults, where Vite fails hard. This landed on main via #135 with your e2e (e2e/config-load-failure.mjs) preserved under your authorship (f660bb0), re-implemented on the current command+mode extraction signatures from deca86c (bdd82b2): extract_vite_values now returns None for an __ok:false payload, adopt_vite_config_values returns Result and fails only when a vite.config is present but broken (an absent config is still fine, and an explicit oj.plugins still takes precedence), and all five call sites propagate it. Closing in favor of the landed version.

PR #41, opened 2026-08-24
fix(plugins): preserve buildStart-only initialization hooks

williamhogman:
Plugins that only implement buildStart are currently discarded while constructing the Vite plugin container. Consequently source generators and initialization plugins never run, even though the bridge explicitly supports buildStart. Preserve these plugins so their initialization runs exactly once. The first commit adds a synthetic regression that fails on upstream main (0 !== 1); the second commit applies the one-line fix. All 104 public unit tests pass (93 passing, 11 fixture-dependent skips). No customer content is included.


raphamorim:
Landed on main as df1ded9 (cherry-picked, authorship preserved). Thanks! The filter now also keeps moduleParsed-only plugins (666a4a3).

PR #43, opened 2026-08-24
fix(plugins): expose the configured Vite mode to plugin hooks

williamhogman:
The Vite plugin bridge hardcodes development or production in each hook context even when the caller explicitly configures staging, preview, or another custom mode. Consequently plugin transformations observe a different mode from the loaded Vite configuration. Preserve the configured mode while retaining the existing production default for builds and development default for serving. This concerns plugin hook contexts, not config-file loading or CLI command selection. The first commit introduces a wholly synthetic regression that fails on upstream main; the second commit provides the minimal fix and protects build defaults. All 104 public unit tests pass (93 passing, 11 fixture-dependent skips). No customer content is included.

### DIFF
--- a/crates/oj_server/src/assets/start/vite-plugin-bridge.mjs
+++ b/crates/oj_server/src/assets/start/vite-plugin-bridge.mjs
@@ -105,7 +105,7 @@ export function findConfig(app) {
   return null;
 }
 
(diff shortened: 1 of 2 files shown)


raphamorim:
Landed on main as 9e15626 (cherry-picked, authorship preserved). Thanks! loadPluginContainer got the same production default for builds in 788fab0.

PR #44, opened 2026-08-24
fix(plugins): pass SSR context to resolve and load hooks

williamhogman:
The plugin bridge passes SSR options to transformation hooks but omits them from resolveId and load. Plugins that provide environment-specific virtual modules therefore return client implementations while preparing the server bundle, breaking SSR compatibility. Pass the Vite-compatible ssr option consistently to both hooks without changing client behavior. The first commit introduces a wholly synthetic regression demonstrating server resolution incorrectly returning a client module; the second commit applies the minimal two-line fix. All 104 public unit tests pass (93 passing, 11 fixture-dependent skips). No customer content is included.


raphamorim:
Landed on main as f031880 (cherry-picked, authorship preserved). Thanks!

PR #45, opened 2026-08-24
fix(config): preserve comma sequences inside JSONC strings

williamhogman:
Summary: Strip JSONC trailing commas only outside quoted strings and correctly tolerate intervening comments. Prevents tsconfig string values from being silently corrupted by comma-plus-bracket sequences. Verification: node --test e2e/unit/loader-util.test.mjs (19 passing; new case fails against upstream main).


raphamorim:
Landed on main as e73c3cf (cherry-picked, authorship preserved). Thanks! Follow-up 2212879 replaces the per-comma slice with a sticky regex lookahead.

PR #46, opened 2026-08-24
fix(ssr): serialize import.meta.glob query objects

williamhogman:
Summary: Serialize object-form import.meta.glob query options into generated import specifiers while preserving original map keys. Adds a synthetic regression covering multiple encoded query parameters. Verification: node --test e2e/unit/glob-transform.test.mjs (8 passing; new case fails against upstream main).


raphamorim:
Landed on main as d22812b (cherry-picked, authorship preserved). Thanks! Matches Vite's stringifyQuery behaviour.

PR #47, opened 2026-08-24
fix(plugins): reset stateful hook filter expressions

williamhogman:
Summary: Reset lastIndex before matching global or sticky plugin-hook filter expressions so repeated modules are consistently transformed. Adds a synthetic regression covering both RegExp modes. Verification: node --test e2e/unit/plugin-gating.test.mjs (12 passing; new case fails against upstream main).


raphamorim:
Already covered on main: matchOne resets lastIndex after every RegExp test (e40abbc), and string id filters are now picomatch-style globs like Vite's pluginFilter, so the substring test here no longer applies. Closing; thanks for the report.

PR #48, opened 2026-08-24
fix(ssr): honor named import.meta.glob exports

williamhogman:
Summary: Preserve Vite named-export selection for eager and lazy import.meta.glob calls instead of returning full module namespaces. Adds a synthetic regression covering both loading modes. Verification: node --test e2e/unit/glob-transform.test.mjs (8 passing; new case fails against upstream main).


raphamorim:
Landed on main as c7b92d8 (cherry-picked, authorship preserved). Thanks!

PR #49, opened 2026-08-24
fix(plugins): honor transform hook source-code filters

williamhogman:
Summary: Apply include and exclude code-content filters to both plugin transform entry points instead of transforming modules that plugins explicitly exclude. Adds a synthetic regression for inclusion, exclusion, and both transform paths. Verification: node --test e2e/unit/plugin-gating.test.mjs (12 passing; new case fails against upstream main).


raphamorim:
Landed on main as ef5c763 (cherry-picked, authorship preserved). Thanks! Follow-up 046b85f adds a dedicated codeAllowed: string id filters became path globs on main, so code filters needed their own substring matcher (Vite's filter.code semantics).

PR #50, opened 2026-08-24
fix(plugins): respect per-hook transform ordering

williamhogman:
Summary: Respect object-form transform hook pre and post ordering in both standard and user-code transform pipelines. Adds a synthetic regression where registration order differs from declared execution order. Verification: node --test e2e/unit/plugin-gating.test.mjs (12 passing; new case fails against upstream main).


raphamorim:
Landed on main as 1f2c2c7 (cherry-picked, authorship preserved). Thanks! Follow-up 046b85f extends per-hook order to resolveId, load, moduleParsed, buildStart, configResolved and generateBundle, as Vite's getSortedPluginHooks does.

PR #51, opened 2026-08-24
fix(ssr): resolve extensionless JSON module imports

williamhogman:
Summary: Include JSON modules and JSON directory indexes in extensionless SSR resolution, matching Vite module resolution. Adds a synthetic regression for both direct modules and index modules. Verification: node --test e2e/unit/loader-util.test.mjs (19 passing; new case fails against upstream main).


raphamorim:
Landed on main as 3f67fd9 (cherry-picked, authorship preserved). Thanks! 2212879 also adds .mts, completing Vite's default resolve.extensions.

PR #52, opened 2026-08-24
fix(plugins): expose SSR options to transform hooks

williamhogman:
Summary: Pass the Vite SSR boolean to transform hooks so environment-aware Markdown and other transforms choose the correct server or client output. Adds a synthetic regression for both environments. Verification: node --test e2e/unit/plugin-gating.test.mjs (12 passing; new case fails against upstream main).


raphamorim:
Landed on main as 745d5e8 (cherry-picked, authorship preserved). Thanks!

PR #53, opened 2026-08-24
fix(plugins): prevent stalled buildStart hooks from blocking SSR

williamhogman:
## Summary
- Give each plugin buildStart hook a bounded startup window so a non-settling hook cannot prevent the SSR bridge becoming ready.
- Preserve normal plugin ordering and continue initializing healthy plugins after a timed-out hook.
- Add independent synthetic regressions for a non-settling initialization hook and ordered successful asynchronous hooks.

## Verification
- Confirmed the synthetic stalled-hook regression fails against the unchanged upstream implementation.
- Confirmed it passes with the timeout fix.
- Ran the JavaScript unit suite successfully.

### DIFF
--- a/crates/oj_server/src/assets/start/vite-plugin-bridge.mjs
+++ b/crates/oj_server/src/assets/start/vite-plugin-bridge.mjs
@@ -105,7 +105,7 @@ export function findConfig(app) {
   return null;
 }
 
-export function createPluginContainer(vite, allPlugins, { command = "serve", mode = "development", environment = "client" } = {}) {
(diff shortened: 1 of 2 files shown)


raphamorim:
Closing without merging: Vite has no buildStart timeout, and a hook that is merely slow (a cold i18n or content compile past 5s) would be skipped and the app would then serve wrong output silently. A buildStart that never settles under oj is a bridge bug to fix at its cause; the new this.resolve/this.load contexts (#63, #65) remove the main reason plugins stalled. Thanks for raising it.

PR #55, opened 2026-08-24
fix(build): respect disabled Vite public directories

williamhogman:
Summary: Preserve publicDir false when loading Vite configuration and skip the public-directory copy during production builds. Adds a synthetic plugin-container regression using a minimal Vite fixture. Verification: node --test e2e/unit/plugin-gating.test.mjs (12 passing; new case fails against upstream main).


raphamorim:
Landed on main as 2dacf8a (cherry-picked, authorship preserved). Thanks!

PR #56, opened 2026-08-24
fix(ssr): match Vite hidden-file glob behavior

williamhogman:
Summary: Exclude hidden files and directories from ordinary import.meta.glob wildcard matches while retaining explicitly requested dot-prefixed patterns. Adds a synthetic regression covering regular files, hidden files, hidden directories, and explicit dot-file matching. Verification: node --test e2e/unit/glob-transform.test.mjs (8 passing; new case fails against upstream main).


raphamorim:
Landed on main as 79b7fb4 (cherry-picked, authorship preserved). Thanks! 2212879 adds the exhaustive option (dot: true and node_modules included), matching Vite's glob options.

PR #57, opened 2026-08-24
fix(config): preserve first-match Vite alias precedence

williamhogman:
Summary: Preserve first-match precedence for duplicate array-form Vite resolve aliases instead of silently overwriting earlier entries. Adds a synthetic conflicting-alias regression. Verification: node --test e2e/unit/vite-extract.test.mjs (8 passing; new case fails against upstream main).


raphamorim:
Landed on main as 81ca40d (cherry-picked, authorship preserved). Thanks!

PR #58, opened 2026-08-24
fix(ssr): omit invalid reserved CommonJS export bindings

williamhogman:
Summary: Exclude strict-mode reserved binding names from generated CommonJS facade named exports so legacy packages cannot produce syntactically invalid ES modules. Adds a synthetic CommonJS fixture with valid and reserved property names. Verification: node --test e2e/unit/loader-util.test.mjs (19 passing; new case fails against upstream main).


raphamorim:
Landed on main as b882d55 (cherry-picked, authorship preserved). Thanks!

PR #59, opened 2026-08-24
fix(config): accept byte-order marks in TypeScript configs

williamhogman:
Summary: Strip an initial UTF-8 byte-order mark before parsing TypeScript JSONC configuration so valid tsconfig files retain their aliases and compiler settings. Adds a synthetic BOM-prefixed configuration regression. Verification: node --test e2e/unit/loader-util.test.mjs (19 passing; new case fails against upstream main).


raphamorim:
Landed on main as c9b018d (cherry-picked, authorship preserved). Thanks!

PR #60, opened 2026-08-24
fix(ssr): resolve fallback package import targets

williamhogman:
Summary: Resolve package import targets expressed as fallback arrays or nested conditional objects, preserving existing import, default, and node condition precedence. Adds synthetic coverage for three valid package import target shapes. Verification: node --test e2e/unit/loader-util.test.mjs (19 passing; new case fails against upstream main).


raphamorim:
Landed on main as 1a905a9 (cherry-picked, authorship preserved). Thanks! 2212879 switches conditional targets to Node's own rule (first supported condition in object order) instead of a fixed import/default/node precedence.

PR #63, opened 2026-08-24
fix(plugins): resolve modules from Vite hook contexts

williamhogman:
Summary: Implement Vite plugin-context module resolution across resolveId, load, transform, buildStart, and generateBundle hooks. Preserve default skipSelf semantics so delegating resolvers cannot recursively reenter themselves. Adds a wholly synthetic virtual-module regression committed before the independent fix. Verification: node --test e2e/unit/plugin-gating.test.mjs (12 passing; new regression fails against upstream main); node --test e2e/unit/asset-routing.test.mjs e2e/unit/rolldown-assets.test.mjs e2e/unit/resolve-pkg.test.mjs (19 passing, 7 optional fixture-dependent skips).

### DIFF
--- a/crates/oj_server/src/assets/start/vite-plugin-bridge.mjs
+++ b/crates/oj_server/src/assets/start/vite-plugin-bridge.mjs
@@ -141,13 +141,23 @@ export function createPluginContainer(vite, allPlugins, { command = "serve", mod
     parse,
   };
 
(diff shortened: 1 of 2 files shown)


raphamorim:
Landed on main as f40f7ad (cherry-picked, authorship preserved). Thanks!

PR #65, opened 2026-08-24
fix(plugins): load modules from Vite hook contexts

williamhogman:
Implement Vite plugin-context module loading in the TanStack Start plugin bridge. Plugins that call this.load with a virtual dependency now receive its module id and source code instead of null, matching the ordinary plugin-host contract and preventing silently swallowed transform failures. This is independent from plugin-context module resolution in PR #63. The wholly synthetic regression is committed separately before the minimal fix. Verification: node --test e2e/unit/plugin-gating.test.mjs (12 passing; regression fails on test-only commit); node --test e2e/unit/asset-routing.test.mjs e2e/unit/rolldown-assets.test.mjs e2e/unit/resolve-pkg.test.mjs (19 passing, 7 optional fixture-dependent skips).


raphamorim:
Landed on main as 33c35cb (cherry-picked, authorship preserved). Thanks!

PR #66, opened 2026-08-24
fix(plugins): expose module graph metadata to hook contexts

williamhogman:
Implement Vite plugin-context module graph inspection in the TanStack Start plugin bridge. Track loaded and transformed modules so plugins can inspect module source and importedIds through getModuleInfo and enumerate observed module identifiers through getModuleIds. The existing public playground already exercises both APIs, but the TanStack bridge previously returned null and an empty iterator unconditionally. This is independent from plugin-context resolution in PR #63 and module loading in PR #65. A wholly synthetic regression is committed separately before the minimal fix. Verification: node --test e2e/unit/plugin-gating.test.mjs (12 passing; new regression fails on its test-only commit); node --test e2e/unit/asset-routing.test.mjs e2e/unit/rolldown-assets.test.mjs e2e/unit/resolve-pkg.test.mjs (19 passing, 7 optional fixture-dependent skips).


raphamorim:
Landed on main as a396205 (cherry-picked, authorship preserved). Thanks! 046b85f adds meta to the module records and shares the record with moduleParsed.

PR #67, opened 2026-08-24
fix(plugins): invoke moduleParsed lifecycle callbacks

williamhogman:
Implement the Vite/Rollup moduleParsed plugin lifecycle in the TanStack Start bridge. Preserve plugins whose only supported hook observes parsed modules and invoke enabled observers with module id, transformed source, and importedIds after both framework and user-code transforms. The public playground graph plugin and its existing browser assertion already require this lifecycle, while the regular plugin host already implements it. This is independent from plugin-context module metadata inspection in PR #66. A wholly synthetic regression is committed before the minimal fix. Verification: node --test e2e/unit/plugin-gating.test.mjs (12 passing; new regression fails against the test-only commit); node --test e2e/unit/asset-routing.test.mjs e2e/unit/rolldown-assets.test.mjs e2e/unit/resolve-pkg.test.mjs (19 passing, 7 optional fixture-dependent skips).


raphamorim:
Landed on main as 666a4a3 (cherry-picked, authorship preserved). Thanks!

PR #68, opened 2026-08-25
fix(plugins): initialize configResolved before module hooks

williamhogman:
Initialize Vite plugin configResolved hooks before TanStack Start plugin bridge module resolution, loading, transformation, build startup, and bundle generation. Share one asynchronous initialization across concurrent module hooks and provide ordinary resolved configuration fields including root, base, command, mode, plugin list, and standard option groups. Existing public playground plugins already initialize closure state in configResolved before consuming it in module hooks. The first commit adds a wholly synthetic regression that fails independently against upstream main; the second commit implements the minimal lifecycle fix. The regression verifies concurrent resolve/load/transform calls observe initialized state and invoke configResolved exactly once. Verification: node --test e2e/unit/plugin-gating.test.mjs e2e/unit/asset-routing.test.mjs e2e/unit/rolldown-assets.test.mjs e2e/unit/resolve-pkg.test.mjs (31 passing, 7 optional fixture-dependent skips). This is independent from plugin config partial merging, moduleParsed callbacks, and module graph metadata.

### DIFF
--- a/crates/oj_server/src/assets/start/vite-plugin-bridge.mjs
+++ b/crates/oj_server/src/assets/start/vite-plugin-bridge.mjs
@@ -105,7 +105,9 @@ export function findConfig(app) {
   return null;
 }
 
-export function createPluginContainer(vite, allPlugins, { command = "serve", mode = "development", environment = "client" } = {}) {
+export function createPluginContainer(vite, allPlugins, {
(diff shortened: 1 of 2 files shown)


raphamorim:
Landed on main as bf23346 (cherry-picked, authorship preserved). Thanks! 788fab0 skips configResolved for the framework plugins oj reimplements (TanStack, Vite internals), which never ran hooks under oj.

PR #69, opened 2026-08-25
fix(plugins): apply Vite config hooks before initialization

williamhogman:
Execute Vite plugin config hooks sequentially before initializing the TanStack Start plugin bridge. Honor command/mode plugin gating and async hooks, preserve nested existing settings, merge partial objects and arrays using vite.mergeConfig when available, and expose updated configuration to later hooks and bridge settings. The public playground config plugin already requires this lifecycle to populate its define values. This fix is independent of configResolved notification: its synthetic regression initializes state exclusively from config hooks and verifies sequential nested/array merging plus configured publicDir. Regression is committed before the minimal implementation. Verification: node --test e2e/unit/plugin-gating.test.mjs (12 passing; regression fails on its test-only commit); node --test e2e/unit/asset-routing.test.mjs e2e/unit/rolldown-assets.test.mjs e2e/unit/resolve-pkg.test.mjs (19 passing, 7 optional fixture-dependent skips).

### DIFF
--- a/crates/oj_server/src/assets/start/vite-plugin-bridge.mjs
+++ b/crates/oj_server/src/assets/start/vite-plugin-bridge.mjs
@@ -241,6 +241,18 @@ export function createPluginContainer(vite, allPlugins, { command = "serve", mod
   return { resolveId, load, transform, transformUserCode, buildStart, generateBundle, pluginCount: plugins.length, watchFiles };
 }
 
+function mergeConfigValues(current, update) {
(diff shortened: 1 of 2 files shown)


raphamorim:
Landed on main as 1cb2559 (cherry-picked, authorship preserved). Thanks! 788fab0 skips the config hooks of plugins oj reimplements (the TanStack router plugin's config hook starts its generator) and gives builds the production default mode; without that the MDX plugin recorded development mode and the production bundle used jsxDEV.

PR #70, opened 2026-08-25
fix(css): resolve transitive Tailwind v4 dependencies

williamhogman:
Resolve Tailwind v4 compiler and scanner dependencies through the existing breadth-first package resolver instead of requiring them to be installed directly at the application root. Strict pnpm-style layouts commonly place @tailwindcss/node and @tailwindcss/oxide beneath the direct @tailwindcss/vite or @tailwindcss/postcss package. Both the development CSS host and production CSS fallback now resolve through those package anchors. A wholly synthetic subprocess regression constructs a strict nested layout, explicitly proves root resolution fails, and verifies CSS compilation still succeeds. The regression is committed before the minimal fix. Verification: node --test e2e/unit/resolve-pkg.test.mjs (12 passing; new strict-layout CSS-host regression fails before fix); node --test e2e/unit/plugin-gating.test.mjs e2e/unit/asset-routing.test.mjs e2e/unit/rolldown-assets.test.mjs (19 passing, 7 optional fixture-dependent skips).

### DIFF
--- a/crates/oj_server/src/assets/start/build.mjs
+++ b/crates/oj_server/src/assets/start/build.mjs
@@ -1,7 +1,6 @@
 // SPDX-License-Identifier: MIT
 
 import { existsSync, readFileSync, writeFileSync, mkdirSync, cpSync, rmSync } from "node:fs";
-import { createRequire } from "node:module";
 import { dirname, join, resolve, relative } from "node:path";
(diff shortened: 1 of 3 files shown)


raphamorim:
Landed on main as 6b3a600 (cherry-picked, authorship preserved). Thanks!

PR #71, opened 2026-08-25
fix(plugins): expose emitted chunks to generateBundle

williamhogman:
Pass actual production client output to Rollup/Vite generateBundle hooks instead of an empty object. This allows existing plugins to inspect generated chunks, emit accurate build reports, and modify entry output; changed chunk source is persisted after hook completion. The public playground report plugin already relies on each of these behaviors. A wholly synthetic fail-first regression passes two generated chunks into the plugin bridge, verifies the emitted report counts both, and confirms the entry chunk is modified. Verification: node --test e2e/unit/plugin-gating.test.mjs (12 passing; synthetic regression fails on test-only commit with chunks:0 instead of chunks:2); node --test e2e/unit/asset-routing.test.mjs e2e/unit/rolldown-assets.test.mjs e2e/unit/resolve-pkg.test.mjs (19 passing, 7 optional fixture-dependent skips).

### DIFF
--- a/crates/oj_server/src/assets/start/build.mjs
+++ b/crates/oj_server/src/assets/start/build.mjs
@@ -190,12 +190,21 @@ const entryChunk = client.output.find((o) => o.type === "chunk" && o.isEntry);
 const clientUrl = "/" + entryChunk.fileName;
 
 if (clientContainer) {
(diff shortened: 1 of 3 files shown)


raphamorim:
Landed on main as 3bac909 (cherry-picked, authorship preserved). Thanks! The Start build reuses the same bundle object for writeBundle (#78).

PR #72, opened 2026-08-25
fix(plugins): expose Vite config resolver to Start hooks

williamhogman:
## Summary
- provide the standard Vite resolved-config resolver to plugins running in the TanStack Start bridge
- honor configured aliases and extensions while reusing the existing strict-layout package resolver for bare imports
- preserve application-provided resolvers and forward the loaded Vite configuration into the bridge

## Regression
- the first commit adds two synthetic regressions that fail independently on upstream main because plugin transforms silently return null when createResolver is absent
- the fix makes both regressions pass, covering aliased TypeScript modules, custom extensions, public packages, missing imports, and existing configured resolvers
- focused regressions: 13 passing; complete public unit suite: 94 passing, 11 skipped because optional fixture dependencies are not installed

### DIFF
--- a/crates/oj_server/src/assets/start/vite-plugin-bridge.mjs
+++ b/crates/oj_server/src/assets/start/vite-plugin-bridge.mjs
@@ -1,8 +1,8 @@
 // SPDX-License-Identifier: MIT
 
-import { importPkg } from "./resolve-pkg.mjs";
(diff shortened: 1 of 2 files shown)


raphamorim:
Landed on main as 4ff7903 (cherry-picked, authorship preserved). Thanks! The resolver now hangs off the resolved config that configResolved (#68) also receives.

PR #73, opened 2026-08-25
fix(plugins): retain configuration-only initializer plugins

williamhogman:
## Summary

- Preserve valid Vite plugins that expose only a configResolved lifecycle callback when the TanStack Start bridge filters its plugin list.
- Keep these initializer plugins alongside operational plugins so multi-plugin integrations do not silently lose their shared initialization component.
- Scope this PR strictly to plugin retention: actual configResolved callback execution is provided separately by #68, and the two changes compose without duplicating its implementation.

## Regression

A synthetic configResolved-only initializer paired with a load-capable consumer produces pluginCount 1 before this change and pluginCount 2 afterward. The regression commit precedes the one-line fix.

## Verification

- Focused regression and related public plugin, asset, Rollup-asset, and package-resolution suites: 31 passing; seven optional fixture-dependent cases skipped.


raphamorim:
Landed on main as 4594388 (cherry-picked, authorship preserved). Thanks!

PR #74, opened 2026-08-25
fix(build): bundle inline HTML module script entries

williamhogman:
## Summary
- recognize standard inline module scripts as production build entry points
- feed their source directly into Rolldown through a virtual plugin without writing temporary application files
- replace inline scripts with bundled asset URLs while preserving configured base paths, script attributes, existing external entry points, and multiple modules

## Regression
- the first commit contains two failing synthetic end-to-end Rust regressions: inline-only pages currently stop with the no module entry error, and mixed pages silently leave inline source unbundled
- the second commit makes both regressions pass, including local imports, multiple inline scripts, a configured base path, preserved nonce and custom attributes, and an existing external module
- complete oj Rust tests: 38 passing; public JavaScript unit suite: 92 passing, 11 skipped because optional fixture dependencies are not installed

### DIFF
--- a/crates/oj/src/build.rs
+++ b/crates/oj/src/build.rs
@@ -2732,6 +2732,83 @@ mod tests {
         assert_eq!(module_script_srcs(html2), vec!["./app/main.ts"]);
     }
 
+    #[tokio::test]
+    async fn build_bundles_inline_module_entries_and_preserves_base_and_attributes() {
+        let suffix = std::time::SystemTime::now()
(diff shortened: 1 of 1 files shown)


raphamorim:
Already on main: inline module scripts are bundled as html-proxy entries since 550d81a (part of the HTML/CSS pipeline work that also bundles stylesheet links). Closing; thanks for the report and the tests, which the html-entries e2e now covers.

PR #76, opened 2026-08-25
fix(plugins): apply production renderChunk hooks

williamhogman:
## Summary

- Retain plugins whose only operational hook is renderChunk in the TanStack Start bridge.
- Chain function-form and object-form hooks in plugin order while honoring environment consumers and exposing the normal plugin context.
- Route hooks through the real Rolldown client renderChunk lifecycle, before generateBundle and output writing, preserving bundler-managed asset hashes and output handling.

OJ already checks in a concrete example: playground/oj.plugins.mjs reportPlugin injects a production entry marker from renderChunk, and the regular plugin host supports that hook, but the Start bridge previously discarded it.

## Regression

A synthetic pre hook, object-form post hook, and server-only hook reproduce the failure: the bridge initially retains zero of three chunk-only plugins. After the fix, client entry code becomes /*client*/entry();/*post*/, the server-only hook stays gated, and non-entry chunks remain unchanged. The failing regression is committed before the implementation.

## Verification

- All public JavaScript unit suites: 93 passing; 11 optional fixture-dependent cases skipped.

### DIFF
--- a/crates/oj_server/src/assets/start/build.mjs
+++ b/crates/oj_server/src/assets/start/build.mjs
@@ -172,6 +172,10 @@ const client = await build({
   resolve: { conditionNames: ["browser", "module", "import"], alias: clientAlias },
   plugins: [
     makeVitePlugins({ container: clientContainer, appRoot: APP, mode: "prod", emit }),
+    {
+      name: "oj-client-render-chunk",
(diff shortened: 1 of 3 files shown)


raphamorim:
Landed on main as 78fe0d6 (cherry-picked, authorship preserved). Thanks! ba17aca wires renderChunk for the server bundle too, since Vite runs it per environment.

PR #77, opened 2026-08-25
fix(build): preserve relative Vite public base paths

williamhogman:
## Summary
- preserve both documented embedded-deployment base values, ./ and the empty string, and expose the matching BASE_URL
- emit relative entry scripts, module preloads, stylesheet links, and CSS image URLs appropriate to each output file
- resolve asynchronously loaded stylesheet URLs against import.meta.url while preserving rooted subpaths and complete CDN URL bases

## Regression
- the first commit adds failing synthetic Rust checks and a true production build fixture showing all generated URLs incorrectly start at the site root
- the fix verifies both documented relative forms, local CSS images, shared-chunk module preloads, multiple entry modules, dynamic stylesheet injection, BASE_URL, nested relative-path calculation, rooted subpaths, and full CDN URLs
- Vite compatibility contract: https://vite.dev/guide/build#relative-base
- complete oj Rust tests: 37 passing; public JavaScript units: 92 passing, 11 optional fixture tests skipped

### DIFF
--- a/crates/oj/src/build.rs
+++ b/crates/oj/src/build.rs
@@ -2273,10 +2273,10 @@ fn lib_format(fmt: &str) -> Option<(&'static str, bool)> {
 
 fn normalize_base(base: &str) -> String {
     if base.is_empty() || base == "./" {
-        return "/".to_string();
+        return "./".to_string();
     }
     let mut b = base.to_string();
-    if !b.starts_with('/') {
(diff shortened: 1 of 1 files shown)


raphamorim:
Already on main: base './' and '' produce page-relative script, modulepreload and stylesheet links, chunk-relative preload deps and BASE_URL './' since 0073a4e (with e2e/preload-relative-base.mjs). Closing; thanks.

PR #78, opened 2026-08-25
fix(plugins): invoke production writeBundle hooks

williamhogman:
## Summary

- Retain standard Rollup/Vite plugins whose only operational hook is writeBundle in the TanStack Start plugin bridge.
- Run function-form and object-form writeBundle hooks in plugin order with the normal plugin context, environment gating, output options, and real emitted client bundle metadata.
- Invoke the post-write lifecycle after generateBundle and the client output has been written.

The checked-in playground already demonstrates the missing behavior: its reportPlugin writes a build report from writeBundle, and the regular OJ plugin host supports the hook, while the TanStack Start bridge previously discarded it.

## Regression

A synthetic pre writer, object-form post writer, server-only writer, and bundle generator reproduce the failure: only one of four plugins survives before the fix. Afterward generation precedes both applicable writers, the server-only hook remains gated, and writers receive actual entry/chunk output plus the client environment. The failing regression is committed before the implementation.

## Verification

- All public JavaScript unit suites: 93 passing; 11 optional fixture-dependent cases skipped.

### DIFF
--- a/crates/oj_server/src/assets/start/build.mjs
+++ b/crates/oj_server/src/assets/start/build.mjs
@@ -196,6 +196,7 @@ if (clientContainer) {
     mkdirSync(dirname(outFile), { recursive: true });
     writeFileSync(outFile, source);
   });
+  await clientContainer.writeBundle(Object.fromEntries(client.output.map((output) => [output.fileName, output])));
 }
 
 const rootManifest = {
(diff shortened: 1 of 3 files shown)


raphamorim:
Landed on main as 6dd9885 (cherry-picked, authorship preserved). Thanks! ba17aca also runs writeBundle for the server bundle.

PR #79, opened 2026-08-25
fix(plugins): expose environment-scoped module graphs

williamhogman:
## Summary
- expose the documented Vite environment module graph to TanStack Start plugin hooks
- support lookup by module ID, asynchronous served URL, and source files shared by queried module variants
- preserve independent client and server module graphs while tracking loaded and transformed source records
- remain independent from Rollup-style plugin-context getModuleInfo and getModuleIds support

## Regression
- the first commit adds two failing synthetic plugin regressions: graph-based transforms currently disappear because the missing environment.moduleGraph throws inside the guarded hook
- the fix verifies loaded module identity, queried source variants, transformed source, missing-module semantics, and client/server isolation across both transform entry points
- official Vite environment API: https://vite.dev/guide/api-environment-plugins
- focused regressions: 13 passing; complete public unit suite: 94 passing, 11 optional fixture tests skipped


raphamorim:
Landed on main as 5dee59b (cherry-picked, authorship preserved). Thanks!

PR #83, opened 2026-08-25
fix(plugins): invoke production closeBundle finalizers

williamhogman:
## Summary
- retain plugins whose only supported lifecycle callback is closeBundle
- invoke function-form and object-form finalizers for the correct client or server environment, continuing past an incompatible callback
- close both production Start plugin containers after output generation and prerendering

## Regression coverage
- committed three synthetic failing regressions before the implementation; upstream main discarded finalizer-only plugins, omitted the container callback, and never finalized production environments
- verified all 33 plugin, package-resolution, and output-asset unit tests pass inside the isolated Lima VM


raphamorim:
Landed on main as 92ebf19 (cherry-picked, authorship preserved). Thanks!

PR #84, opened 2026-08-25
fix(build): honor Vite module preload configuration

williamhogman:
## Summary
- preserve Vite `build.modulePreload` from both `oj.config.json` and `vite.config.mjs`
- skip generated modulepreload links only when the option is explicitly `false`
- retain default preloading and object-form settings such as `{ polyfill: false }`

## Regression
The first commit adds a synthetic shared-chunk build test that fails on main because `modulePreload: false` still injects `<link rel="modulepreload">`. The second commit makes all four configuration cases pass.

## Verification
- All 92 available JavaScript unit tests passed inside the VM; 11 dependency-backed cases were skipped because their optional fixtures are not installed.

### DIFF
--- a/crates/oj/src/build.rs
+++ b/crates/oj/src/build.rs
@@ -2599,6 +2599,67 @@ fn build_manifest(entries: &[ManifestEntry]) -> serde_json::Value {
 mod tests {
     use super::*;
 
+    #[tokio::test]
+    async fn module_preload_configuration_controls_production_links() {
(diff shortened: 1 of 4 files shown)


raphamorim:
Landed on main as e7395f2 (cherry-picked, authorship preserved). Thanks! main already carried build.modulePreload for the polyfill, so the commit was trimmed to the link gate (Vite's html plugin skips modulepreload links only when modulePreload === false) plus your test.

PR #85, opened 2026-08-25
fix(plugins): invoke Rollup buildEnd lifecycle hooks

williamhogman:
## Summary
- retain plugins whose only supported Rollup lifecycle callback is buildEnd
- invoke function-form and object-form build completion hooks only for their matching environment while preserving the original build error
- forward Rolldown buildEnd callbacks exclusively to the active Vite plugin container

## Regression coverage
- committed three synthetic failing regressions before the implementation; upstream main discarded buildEnd-only plugins, omitted the container callback, and never forwarded Rolldown completion
- verified all 22 plugin and output-asset unit tests pass inside a separate isolated Lima VM directory


raphamorim:
Landed on main as f06f881 (cherry-picked, authorship preserved). Thanks!

PR #86, opened 2026-08-25
fix(build): respect Vite output directory cleanup setting

williamhogman:
## Summary
- preserve `build.emptyOutDir` from both `oj.config.json` and `vite.config.mjs`
- retain existing output files when `emptyOutDir: false` is explicitly configured
- continue clearing previous output by default and when `emptyOutDir: true`

## Regression
The first commit adds a synthetic production-build test that fails on main because OJ always deletes the output directory. The second commit makes direct configuration, Vite configuration, explicitly enabled cleanup, and default cleanup all pass.

## Verification
- All 92 available JavaScript unit tests passed inside the VM; 11 optional dependency-backed cases were skipped.

### DIFF
--- a/crates/oj/src/build.rs
+++ b/crates/oj/src/build.rs
@@ -1111,7 +1111,9 @@ pub async fn build(
 
     let base = normalize_base(config.base.as_deref().unwrap_or("/"));
 
-    let _ = fs::remove_dir_all(&out_dir);
+    if build_cfg.empty_out_dir.unwrap_or(true) {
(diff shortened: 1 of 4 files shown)


raphamorim:
Already on main: emptyOutDir follows Vite's rule since 46b75a3 (an outDir outside root is left alone unless build.emptyOutDir or --emptyOutDir says so, .git is preserved), from both oj.config and vite.config. Closing; thanks.

PR #87, opened 2026-08-25
fix(plugins): invoke Rollup renderStart lifecycle hooks

williamhogman:
## Summary
- retain plugins whose only supported Rollup lifecycle callback is renderStart
- invoke function-form and object-form render initialization hooks only for their matching environment while preserving both output and input options
- forward Rolldown renderStart callbacks exclusively to the active Vite plugin container

## Regression coverage
- committed three synthetic failing regressions before the implementation; upstream main discarded renderStart-only plugins, omitted the container callback, and never forwarded Rolldown render initialization
- verified all 22 plugin and output-asset unit tests pass inside a separate isolated Lima VM directory


raphamorim:
Landed on main as da4082e (cherry-picked, authorship preserved). Thanks!

PR #88, opened 2026-08-25
fix(build): honor Vite public directory copy setting

williamhogman:
## Summary
- preserve `build.copyPublicDir` from both `oj.config.json` and `vite.config.mjs`
- skip copying `public/` when `copyPublicDir: false` is explicitly configured
- continue copying public assets by default and when `copyPublicDir: true`

## Regression
The first commit adds a synthetic production-build test that fails on main because OJ copies public files even with `copyPublicDir: false`. The second commit makes direct configuration, Vite configuration, explicit copying, and default copying all pass.

## Verification
- All 92 available JavaScript unit tests passed inside the VM; 11 optional dependency-backed cases were skipped.

### DIFF
--- a/crates/oj/src/build.rs
+++ b/crates/oj/src/build.rs
@@ -1440,7 +1440,9 @@ pub async fn build(
         .as_ref()
         .map(|p| root.join(p))
         .unwrap_or_else(|| root.join("public"));
-    copy_public_dir(&public_dir, &out_dir)?;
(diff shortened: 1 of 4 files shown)


raphamorim:
Landed on main as 121f350 (cherry-picked, authorship preserved). Thanks! b6a8bb9 folds the extraction into the shared build block that the other build options use.

PR #110, opened 2026-08-28
accept an explicit config path on oj build

mikn:
`oj dev` takes `--config <PATH>`. `oj build` takes only `[ROOT] --out --ssr
--mode`, so the only way to tell a build which config to read is to put the file
where the ROOT-relative convention search looks. A tool that generates its
config outside the app tree has nothing to put there, and handing the path over
positionally makes oj read it as ROOT.

Both commands already share the mechanism. `set_vite_config_override` pins what
`vite_config_file` returns, and `build()` canonicalizes ROOT and then calls
`adopt_vite_config_values` in the same order `DevServer::build_app` does. The
flag goes into that same call and resolves a relative path against the canonical
ROOT the way dev does, so `dev` and `build` cannot disagree about which file is
the config. The added block in `crates/oj/src/build.rs` is the same shape as the
one already in `crates/oj_server/src/lib.rs`.

The TanStack Start branch is untouched, which is what dev does too: there the
flag reaches `DevServer` and not `start_dev`.

## Testing

`e2e/config-flag.mjs` was already the `--config` fixture — it builds an app with
a root `vite.config.mjs` plus an override config and asserts the override's
`transformIndexHtml` plugin fires only with `--config`. It now does the same for
`build`, reusing that fixture: one build with the root config asserts the
override marker is absent, one with `--config` asserts it is present. Without the
flag clap rejects the argument and the test throws; with the flag parsed but
inert the second assertion fails.

- `node e2e/config-flag.mjs` — passes
- `cargo test --workspace` — green
- `cargo build -p oj` — clean
- `cargo clippy -p oj` — no new warnings (`too_many_arguments` does not fire;
  `build()` is 5 args against a threshold of 7)

## One request

Please don't fold a `cargo fmt --all` into this branch. The checked-in tree is
not fmt-clean against rustfmt 1.95 (the version `rust-toolchain.toml` pins):
running it rewrites about 1100 lines across 37 files of unrelated code. The lines
added here are fmt-clean on their own — `rustfmt --check` reports no hunk
touching `config_path` or `set_vite_config_override`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@ All notable changes to oj are documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Added
+- `oj build --config <path>` names the config to build with, the same flag `oj dev` already takes. Without it the config could only be found by convention relative to the positional root, so a config generated outside the app tree was unreachable.
+
 ## [0.1.6] - 2026-08-26
 
 ### Fixed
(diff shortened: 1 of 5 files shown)


raphamorim:
Already on main: oj build (and preview) take --config since 1f2c2c7, resolved against the app root like dev; e2e/build-config-flag.mjs covers it. Closing; thanks for the write-up, it matched the fix exactly.

PR #116, opened 2026-08-30
fix(start): let tsr.config.json configure the route generator

mikn:
## The failure

An app whose route tree is not at the default path gets a second one it never
asked for. `oj dev` writes `src/routeTree.gen.ts`; the app goes on importing the
tree it configured; the two drift apart immediately, and for an app that checks
its route tree in, every `oj dev` dirties the working tree.

## Why

`generate.mjs` passes its settings to `getConfig` **inline**, and
`@tanstack/router-generator` merges them the way that makes inline win:

```js
const merged = { ...fileConfigRaw, ...inlineConfig };
```

So `routesDirectory`, `generatedRouteTree`, `autoCodeSplitting` and
`routeFileIgnorePattern` — every one of them a documented `tsr.config.json`
setting — cannot be set by the app at all. `routeTreeFileFooter` is in the same
schema and equally unreachable.

## The fix

Only `target` is genuinely the adapter's to decide. The other four become
*defaults*: supplied when the app has not set them, dropped when it has. An app
with no `tsr.config.json` resolves exactly as before.

## Tests

`e2e/unit/start-route-generator.test.mjs`, skipping when the `start-app` fixture
has no `node_modules`, following the convention in `optimize-deps.test.mjs`:

- the default path still applies when the app configures nothing — this is the
  no-regression half, and it passes on `main` too;
- `tsr.config.json` decides where the tree is written, and the default path does
  **not** also get one. This fails on `main` and passes here.

Verified by reverting `generate.mjs` and re-running: the first test passes, the
second fails.

## How I hit it

to find where the two diverge. The app in question keeps its tree at
`src/routes/routeTree.gen.ts` so the tree and the routes it types sit in one
compilation unit.

### DIFF
--- a/crates/oj_server/src/assets/start/generate.mjs
+++ b/crates/oj_server/src/assets/start/generate.mjs
@@ -1,5 +1,8 @@
 // SPDX-License-Identifier: MIT
 
+import { readFileSync } from "node:fs";
+import { join as pathJoin } from "node:path";
+
 import { importPkg } from "./resolve-pkg.mjs";
 
 const root = process.env.OJ_APP_ROOT ?? process.cwd();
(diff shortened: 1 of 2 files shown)


raphamorim:
Landed on main as e080844 (cherry-picked, authorship preserved). Thanks! The declare-module footer (#118) now follows the configured generatedRouteTree too.

PR #117, opened 2026-08-30
fix(start): let package.json "imports" override the framework-seam conventions

mikn:
## The failure

An app that configures its router entry builds under Vite and fails under `oj`:

```
[RESOLVE_ERROR] Could not resolve '#tanstack-router-entry' in
  .../@tanstack/start-client-core/dist/esm/client/hydrateStart.js
  Matched alias not found for '#tanstack-router-entry'
```

`tanstackStart({ router: { entry: "lib/router" } })` is a supported framework
option — `start-plugin-core/plugin.js` resolves it through `resolveEntry` with
`defaultEntry: "router"` against `srcDirectory`, then aliases
`#tanstack-router-entry` at the resolved file.

## Why

The adapter hardcodes `<app>/src/router` — the framework default — in three
places:

| site | what it feeds |
| --- | --- |
| `loader.mjs` `ALIASES` | SSR resolution |
| `bundle-client.mjs` `routerEntry()` | rolldown `resolve.alias`, dev client bundle |
| `build.mjs` `routerEntry()` | rolldown `resolve.alias`, production build |

## The fix

`loader.mjs` already implements package.json `imports`, and an app declaring
`"#tanstack-router-entry"` there is the standard way to name a subpath. The only
problem is precedence: `ALIASES` is consulted first, so a *declaration* loses to
a *convention*. This inverts those two for `#`-prefixed specifiers, in the module
path and the asset path, and gives the two `routerEntry()`s the same rule.

An app with no `imports` entry resolves exactly as before.

## What this is not

The complete fix is for the adapter to read the app's Start configuration —
that would settle `srcDirectory` and the client/server/start entries too, not
just the router. But `start_dev(root, port, host)` takes no config and builds
its `DevServer` with `config: None`, so the app's Vite config never reaches this
code at all. I'd be glad to attempt that separately if you'd want it; this is the
smaller half that unblocks an app today.

## Tests

`e2e/unit/start-router-entry.test.mjs`, skipping when the `start-app` fixture
has no `node_modules`. The fixture app declares its entry at `src/lib/router.ts`
**and** has a `src/router.ts`, so a pass cannot come from the convention
happening to find the same file. Fails on `main`, passes here.

## How I hit it

to find where the two diverge.

### DIFF
--- a/crates/oj_server/src/assets/start/build.mjs
+++ b/crates/oj_server/src/assets/start/build.mjs
@@ -63,7 +63,24 @@ const SERVER = [
   '',
 ].join("\n");
 
+// The app's package.json "imports" wins over the src/router convention, the
+// way it does in the SSR loader: the framework makes this path configurable
+// (`router.entry`), and an app that moved it has to be able to say so.
+function declaredRouterEntry() {
+  try {
+    const target = JSON.parse(readFileSync(resolve(APP, "package.json"), "utf8"))
+      .imports?.["#tanstack-router-entry"];
+    if (typeof target !== "string") return null;
+    const p = resolve(APP, target);
+    return existsSync(p) ? p : null;
+  } catch {
+    return null;
+  }
+}
+
 function routerEntry() {
+  const declared = declaredRouterEntry();
+  if (declared) return declared;
   for (const ext of [".tsx", ".ts", ".jsx", ".js"]) {
     const p = resolve(APP, "src/router" + ext);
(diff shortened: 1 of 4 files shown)


raphamorim:
Landed on main as 083eb0d (cherry-picked, authorship preserved). Thanks! The loader's resolution had been refactored into resolveSpec on main; the imports-before-aliases order is applied there.

PR #118, opened 2026-08-30
fix(start): emit TanStack Start's declare-module footer from the generator

mikn:
## The failure

`oj dev` rewrites a checked-in `routeTree.gen.ts` on every boot, stripping the
`declare module` block the framework's own tooling puts there. The tree the
adapter generates and the tree `@tanstack/react-start`'s Vite plugin generates
are different files.

## Why

`start-plugin-core`'s `start-router-plugin` passes a `routeTreeFileFooter` to
the generator, built by `moduleDeclaration()`:

```ts
import type { getRouter } from './router.ts'
import type { createStart } from '@tanstack/react-start'
declare module '@tanstack/react-start' {
  interface Register {
    ssr: true
    router: Awaited<ReturnType<typeof getRouter>>
  }
}
```

`Register` is what types every `createFileRoute()` in the app. `generate.mjs`
passes no footer at all, so the block is absent from anything it writes.

This matters most when the type-check is a separate step from the bundler, so
the tree is checked in and guarded — then the adapter and the framework fight
over the file.

## The fix

Build the footer the same way the plugin does: the router entry relative to the
generated tree's directory, POSIX-separated, `./`-prefixed when needed. The
no-start-entry form is emitted, matching the adapter's synthesized start entry.

The router entry is resolved by the `src/router` convention **and** by the app's
`package.json` `imports` when it has one — the same two sources the seam is
resolved from elsewhere, so the footer cannot name a different file from the one
`#tanstack-router-entry` actually resolves to.

## Tests

`e2e/unit/start-route-tree-footer.test.mjs`, skipping when the `start-app`
fixture has no `node_modules` (the convention `optimize-deps.test.mjs` uses):
the footer is present and well-formed, and it follows a declared router entry.
Both fail on `main` and pass here.

## Relationship to my other PRs

Independent of #115, #116 and #117 — this applies to `main` on its own. If #116
(tsr.config.json) lands first there is a small textual conflict in the same
`getConfig` call, which I'll happily rebase; the footer would then want the
*configured* tree path rather than the constant, and I can follow up with that.

## How I hit it

to find where the two diverge.

### DIFF
--- a/crates/oj_server/src/assets/start/generate.mjs
+++ b/crates/oj_server/src/assets/start/generate.mjs
@@ -1,5 +1,8 @@
 // SPDX-License-Identifier: MIT
 
+import { existsSync, readFileSync } from "node:fs";
+import { dirname, join as pathJoin, relative as pathRelative, resolve as pathResolve, sep } from "node:path";
+
 import { importPkg } from "./resolve-pkg.mjs";
 
 const root = process.env.OJ_APP_ROOT ?? process.cwd();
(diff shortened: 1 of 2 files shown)


raphamorim:
Landed on main as ced30ed (cherry-picked, authorship preserved). Thanks! Two adjustments: the footer's relative import follows the tree path tsr.config.json configures (#116), and an app that sets routeTreeFileFooter itself keeps its own.

PR #119, opened 2026-08-30
fix(start): thread --config through to the TanStack Start dev server

mikn:
## The failure

`oj dev --config <PATH> <ROOT>` is accepted, and then dropped, when the app is a
TanStack Start app.

The one you notice first is the port. A config that sets `server.port` is
ignored, so oj serves on its default 5199 with no diagnostic — the caller thinks
it asked for a port and got one:

```
$ oj dev --config /out/vite.config.mjs /srv/app     # config sets port 5174
  oj dev (tanstack start)
  http://localhost:5199/
```

Everything else the config carries goes the same way: `server.fs.allow`,
`resolve.alias`, `resolve.dedupe`, `define`, and the app's own plugins.

## Why

```rust
// crates/oj/src/main.rs
} else if oj_server::is_tanstack_start_app(&root) {
    start_dev::start_dev(root, port, host).await     // `config` is in scope, and not passed
}
```

`start_dev` takes only `(root, port, host)` and constructs its `DevServer` with
`config: None`. Both halves then fall back to the convention search for
`vite.config.{ts,mts,mjs,js}` *inside the served root*, so a config anywhere else
is invisible. That is the normal case for a build system that generates one — it
has an output tree, not a spot in the app.

The non-Start path already threads it: `DevServer { config, .. }`.

## The fix

Threaded exactly the way `build` does it — resolve a relative path against the
app root, pin `set_vite_config_override` before anything reads a config, then
hand the same path to `DevServer`. So `oj dev --config X app` and
`oj build --config X app` cannot disagree about which file is the config.

The path resolution is pulled into a `config_path` helper with a unit test
beside the other private path helpers in `start_dev.rs`, which is where
`document_url` and `percent_decode` live.

## How I hit it

Building Bazel rules that run `oj dev` and `vite dev` from one generated config.
The config is generated into the build's output tree, so the convention search
can never find it; the port being silently 5199 is what made it visible.

Related to my other Start PRs (#115, #116, #117, #118) but independent of them —
this one applies to `main` on its own.

### DIFF
--- a/crates/oj/src/main.rs
+++ b/crates/oj/src/main.rs
@@ -106,7 +106,7 @@ async fn run() -> anyhow::Result<()> {
             if let Some(entry) = ssr {
                 ssr_dev::ssr_dev(root, entry, port, host).await
             } else if oj_server::is_tanstack_start_app(&root) {
-                start_dev::start_dev(root, port, host).await
+                start_dev::start_dev(root, port, host, config).await
             } else {
                 oj_server::DevServer {
                     root,
--- a/crates/oj/src/start_dev.rs
+++ b/crates/oj/src/start_dev.rs
@@ -37,14 +37,29 @@ struct StartState {
     css_host: Option<Arc<tokio::sync::Mutex<Runner>>>,
 }
 
+// A --config path is resolved against the app root, the way `build` resolves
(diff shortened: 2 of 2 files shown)


raphamorim:
Landed on main as fc06341 (cherry-picked, authorship preserved). Thanks!

PR #120, opened 2026-08-30
feat(cache): let OJ_CACHE_DIR relocate the cache root

mikn:
## What

`oj_cache::cache_root` is `<app>/.oj-cache` with no way to move it, so serving
an app writes into the app.

That is fine for a checkout you own. It is wrong for a source tree you do not:

- **read-only or shared** — a build system may serve a source tree it did not
  create, or one several concurrent builds share;
- **checked for cleanliness** — a tree that is expected to be unmodified after a
  command runs;
- **walked by the build's own tooling** — in my case Gazelle generated a
  `BUILD.bazel` per cached blob under `.oj-cache/v1/**` until the directory was
  explicitly ignored, which is a fairly loud symptom of an artefact landing
  somewhere it is not expected.

`OJ_SSR_BRIDGE_DIR` already lets the SSR bridge move for the same reason, so
this follows it rather than inventing a convention.

## Details

- Unset → unchanged: `<app>/.oj-cache/v<N>`.
- Set → `$OJ_CACHE_DIR/v<N>`.
- Empty → falls back. An unset and a blank environment variable mean the same
  thing to most shells that construct one, and `/v1` would be a bad answer.

The version segment is appended either way: it is what makes an older layout
inert rather than misread, and a caller pointing one directory at several apps
still gets that.

## Tests

`cache_root_is_relocatable_and_stays_versioned` in `oj_cache`, covering the
three cases above. The existing `cache_root_is_versioned` still passes. The env
is process-wide, so the test restores whatever it found.

## How I hit it

Building Bazel rules that run `oj dev`. Bazel's model is that a build writes to
its own output tree and leaves the source tree alone, so `.oj-cache` appearing
in the workspace is the one artefact I could not keep out of it.

### DIFF
--- a/crates/oj_cache/src/lib.rs
+++ b/crates/oj_cache/src/lib.rs
@@ -17,10 +17,30 @@ pub const CACHE_FORMAT: u32 = 5;
 
 pub const CACHE_ROOT_VERSION: u32 = 1;
 
+/// Where everything oj caches for an app goes.
+///
+/// `<app>/.oj-cache` by default, and `OJ_CACHE_DIR` when the caller has
+/// somewhere else for it. A build system serving a source tree it does not own
+/// -- read-only, shared between concurrent builds, or checked for cleanliness
+/// afterwards -- needs the writes to land outside it, and `OJ_SSR_BRIDGE_DIR`
+/// already works this way for the SSR bridge.
+///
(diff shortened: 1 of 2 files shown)


raphamorim:
Landed on main as 9d98e8d (cherry-picked, authorship preserved). Thanks!

PR #121, opened 2026-08-30
fix(config): say when a vite config produced no values

mikn:
## What

`extract_vite_values` has three ways to decline, and none of them says anything.
The server starts, serves, and has `server.fs.allow`, `resolve.alias`, `define`,
`dedupe` and `optimizeDeps` quietly absent — which looks exactly like a config
that had none of those in it.

| path | what happens today |
| --- | --- |
| `--config` names a file that is not readable | `vite_config_file` → `None`, so does this |
| the extractor writes something that is not JSON | `from_slice(...).ok()?` swallows it |
| the extractor writes nothing at all and exits 0 | `{}`-shaped absence, indistinguishable from an empty config |

## Why the third one matters

That is what a run looks like when the extractor skips its own body. It is
reachable today: `vite-extract.mjs` guards everything behind

```js
const isMainRun = import.meta.url === pathToFileURL(process.argv[1] || "").href;
```

Node canonicalises the entry module, so `import.meta.url` is symlink-free while
`argv[1]` is whatever the caller passed. Any symlink in that path makes them
disagree — every path under `/var` on macOS arrives as `/private/var` — and the
script then does nothing, writes nothing, and exits 0.

I hit that for real: `server.fs.allow` and `resolve.alias` silently stopped
being applied on macOS only, and it took four rounds of bisecting CI to find,
because nothing anywhere said the config had not been read. With this patch it
is one line:

```
oj: extracting /…/vite.config.mjs wrote nothing at all and exited with exit status: 0;
server.fs.allow, resolve.alias, define and the rest of that config are not in effect
```

I have a separate one-line fix for the `isMainRun` comparison itself (compare
realpaths) and am happy to send it — but the silence is the more valuable fix,
because it is what let a two-line bug hide.

## Tests

The classification is pulled into `extraction_failure` so it can be tested
without provoking a real `node` into failing three different ways:
`valid_json_is_not_a_failure`, `a_silent_successful_run_is_a_failure`,
`unparseable_output_reports_its_size_and_the_parse_error`.

## How I hit it

which means the config lives in a build output directory rather than beside the
app — so the extractor's path is one I construct, and it had a symlink in it.

### DIFF
--- a/crates/oj_server/src/plugins.rs
+++ b/crates/oj_server/src/plugins.rs
@@ -192,11 +192,46 @@ pub struct ViteValues {
     pub optimize_deps: Option<serde_json::Value>,
 }
 
+/// Why a run of the extractor produced nothing usable, or None when it did.
+///
+/// Kept separate from the reporting so the classification can be tested: the
+/// interesting cases are a subprocess that wrote nothing at all and one that
+/// wrote something that is not JSON, and reproducing either through a real
+/// `node` is harder than it is worth.
+fn extraction_failure(status: std::process::ExitStatus, stdout: &[u8], parse: Option<&str>) -> Option<String> {
+    match parse {
+        None => None,
+        Some(_) if stdout.is_empty() => Some(format!(
+            "wrote nothing at all and exited with {status}"
+        )),
+        Some(e) => Some(format!(
+            "wrote {} bytes that are not JSON ({e}) and exited with {status}",
+            stdout.len()
+        )),
+    }
+}
+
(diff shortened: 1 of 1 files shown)


raphamorim:
Landed on main as a04725e (cherry-picked, authorship preserved). Thanks! main had meanwhile made a present-but-unloadable vite.config a hard error, so your failure classification now feeds that error's diagnostic, and a --config that names a missing file is an error as well rather than a warning (e2e/config-flag.mjs covers it).

PR #123, opened 2026-08-30
fix(config): decide isMainRun on the real path, not the spelling

mikn:
## The bug

`vite-extract.mjs` guards its entire body with:

```js
const isMainRun = import.meta.url === pathToFileURL(process.argv[1] || "").href;
if (isMainRun) try { … }
```

Node canonicalises the entry module, so `import.meta.url` is symlink-free. `argv[1]`
is whatever the caller passed. **Any symlink in that path makes the two disagree** —
and on macOS every path under `/var` arrives as `/private/var`, so it is not an
exotic case.

When they disagree the script does nothing, writes nothing, and exits 0.

## Reproduction

Two commands, on any platform:

```console
$ mkdir -p /tmp/x/real && ln -s /tmp/x/real /tmp/x/link
$ cp crates/oj_server/src/assets/vite-extract.mjs /tmp/x/real/

$ node /tmp/x/real/vite-extract.mjs /nonexistent.mjs /tmp
oj: could not extract vite.config values: Error [ERR_MODULE_NOT_FOUND] …    # runs

$ node /tmp/x/link/vite-extract.mjs /nonexistent.mjs /tmp
$                                                                          # exit 0, nothing
```

## Why it matters in practice

`extract_vite_values` builds the script path from `cache_root(root)`. Today
`root` has already been canonicalised by `build_app`, so the paths happen to
agree and nobody notices. They stop agreeing as soon as the cache root comes
from anywhere else — which is exactly what happens with `OJ_CACHE_DIR`
(#120), or for anyone pointing oj at a path that goes through a symlink.

What I actually saw: `server.fs.allow` and `resolve.alias` silently stopped
being applied on macOS only. `/@fs` answered 403 and a first-party bare
specifier resolved nowhere, with no error anywhere — because a skipped body and
an empty config are the same thing to the caller. It took four rounds of
bisecting CI to find.

## The fix

Compare realpaths, falling back to the resolved path when a component does not
exist so a missing entry is still a mismatch rather than a throw.

## Tests

`e2e/unit/vite-extract-main-run.test.mjs` runs the extractor through both a real
directory and a symlink to it and requires the same output from each. Fails on
`main`, passes here. Full `e2e/unit` suite: 125 pass.

## Related

#121 makes this class of failure audible — it reports when an extraction
produced no values instead of returning an all-`None` config. That patch is what
found this one, and the two are independent: this fixes the cause, #121 makes
the next one like it visible in a single log line.

### DIFF
--- a/crates/oj_server/src/assets/vite-extract.mjs
+++ b/crates/oj_server/src/assets/vite-extract.mjs
@@ -4,7 +4,7 @@
 import { createRequire } from "node:module";
 import { pathToFileURL, fileURLToPath } from "node:url";
 import { dirname, resolve } from "node:path";
-import { writeFileSync, readFileSync } from "node:fs";
+import { writeFileSync, readFileSync, realpathSync } from "node:fs";
 
 const configPath = process.argv[2];
 const appRoot = process.argv[3];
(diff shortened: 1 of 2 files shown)


raphamorim:
Landed on main as 141768a (cherry-picked, authorship preserved). Thanks!

Issue #146, opened 2026-09-03
Rebuild loop, vite.config load failure, and 'document is not defined' in SSR on a large TanStack Start app (0.1.13, linux)


raphamorim:
On a large TanStack Start app (~14k modules, 52 vite plugins) that serves correctly under oj 0.1.8 on macOS, oj 0.1.13 on linux inside a container shows three co-occurring failures:

1. **Continuous rebuild loop** (~6 "rebuilt, reloading" cycles per minute) when the worktree has unrelated file churn — possibly missing watcher ignores for cache/log/state paths (`.oj-cache/`, tool state dirs, log files).
2. **`failed to load config from vite.config.ts`** — config eval fails in this environment; the same file loads fine on macOS under 0.1.8.
3. **`ReferenceError: document is not defined`** during SSR warmup, plus SSR warmup fetch failures — browser-only code executing server-side. Worth checking whether a client-profile transform (client defines inline `SSR: false`) can cross into the SSR pipeline via a shared module/transform cache entry, which would run `if (!import.meta.env.SSR)` branches on the server.

Suspect surface given the delta: changes in 0.1.9–0.1.13, linux-specific behavior, or busy-worktree interaction — not the app config itself. The three may be one cascade (config-load failure at the root, watcher and SSR misbehavior downstream).

Happy to bisect versions or add debug output on request.

raphamorim:
All three symptoms are root-caused and fixed as of 0.1.17/0.1.18 — and they were indeed one cascade plus two independents:

**2. `failed to load config from vite.config.ts`** had two layers. The trigger was a boot-time race: the route tree, server-fn resolver, config extractor and plugin host all bundle `vite.config.ts` concurrently, and the extractor script plus the bundled config shared fixed filenames, so one process could truncate a file another was importing — busy checkouts (exactly the reported environment) made it likely. Fixed in 0.1.17: the script lands by atomic rename, every bundle gets a unique temp name, and the extractor returns JSON through a per-call file instead of stdout (which plugin code, e.g. a route generator's warnings, could pollute). The latent second layer — the esbuild fallback loader keeping bare specifiers that don't resolve from the cache dir when a config inlines a sibling workspace package's source — is fixed in 0.1.18 (#155): both config bundlers now mirror Vite's `externalize-deps` (importer-resolved absolute externals, first-party/TS sources inlined, Node resolution conditions, per-file `__dirname`/`import.meta.url` scope variables).

[...]

PR #147, opened 2026-09-03
Node SSR loader: strip the browser condition and key the resolve cache on conditions

raphamorim:
Fixes the `document is not defined` SSR failure in #146 (root-caused; the other two signatures there are likely downstream).


raphamorim:
Second review pass against Vite's source found and fixed a gap in the external split: `isExternalizable` has an `isInNodeModules(resolved.id)` leg, a bare import whose package realpaths **outside** node_modules (linked/workspace package) is NOT externalized and keeps the environment's conditions. The spec-only gate missed that, giving workspace deps externalConditions.

PR #148, opened 2026-09-03
Native plugin seam: OjPlugin trait, compiler slot, dev and build hosts

raphamorim:
Phase 1 of a native plugin system for oj: the generic seam, with no first-party plugin yet. Design per the plugin-system memo (two tiers: compile-time Rust plugins plus the existing JS bridge; no dynamic libraries, no WebAssembly). Acceptance bar: with no native plugin registered, behavior is byte-identical and the whole existing suite passes unchanged.

## What lands
- **`oj_plugin` crate**: `OjPlugin: rolldown_plugin::Plugin` with defaulted methods (`cache_salt`, `replaces_js_plugins`, `config_section`, `transform_filter`, `pre_transform_ast`, `mtime_cacheable`, `module_seen`, `module_removed`, `css_directive`, `css_for_directive`, `virtual_css`, `css_parse_options`, `invalidates`), an object-safe view, a `Registry`, and `ModuleFilter` over rolldown's own hook-filter interpreter (id, code and module-type includes and excludes, evaluated before parsing).
- **Compiler slot**: `oj_compiler::compile_module_full` runs registered pre-transformers on the oxc `Program` between the parser and the semantic build, so passes see JSX and TypeScript; side channels come back as `CompileOutput.meta`.
- **Dev host**: registry on the server; the slot is filter-gated in `ensure_module`; `module_seen` runs on cold compiles and on every cache-hit path (mtime, memory, persistent) so registries survive restarts; `module_removed` on deletion; `invalidates` in the HMR decision (css-update for a url, module change, full reload); a plugin `@directive;` is masked through the sidecars with an oj-owned sentinel and expanded after Lightning CSS; `virtual_css` is served at `/@oj/<name>.css`.
- **Build adapter**: one rolldown plugin parses a wanted module once, runs the same pass, prints with a source map; virtual sheets become `assets/<name>-<hash>.css` with a manifest row and a page link; the directive uses the same sentinel in `compile_stylesheet`.
- **Cache and config**: `CachedModule.meta` (omitted when empty; older entries load), the persistent-cache salt folds every active plugin's `cache_salt`, `mtime_cacheable` gates the fast path; `OjConfig` gains a flatten catch-all so `config_section(name)` reads a plugin's own section; `replaces_js_plugins` is unioned into the native-plugin skip list handed to the JS host.
- **Proof plugin**: `example-marker` behind the default-off `example-plugin` cargo feature (active only with a `marker` config section or `OJ_EXAMPLE_PLUGIN=1`) exercises both archetypes the memo identified: an AST pass with a side channel, and a cross-module registry feeding a virtual sheet and a directive with css-update invalidation.

## Tests
36 new Rust unit tests (filters, registry, cache salt fold and meta round trip, config catch-all, directive sentinel through Lightning CSS, the example plugin), a plugin-host unit test for the skip-list union, `e2e/native-plugin-dev.mjs` (rewrite, virtual sheet, directive, css-update on change, warm-restart replay of `module_seen` from cached meta) and `e2e/native-plugin-build.mjs` (rewritten chunk, emitted sheet, manifest row, link). Full workspace tests, JS unit tests, run.mjs in both modes, start.mjs and every CI e2e pass.

## Deferred (documented in code and in the follow-up list)
Migrating the first builtin (the `import.meta.glob` family needs an after-transform position and a "no source map" flag), a dev `PluginContext` for rolldown hooks, a consumer for `css_parse_options`, and the cache-salt field-flip test helper.


raphamorim:
Review fixes pushed in the last commit:

- **Build: directives expanded after the bundle.** `compile_stylesheet` now only masks; collected sheets are expanded in `unmask_collected_css` right after every `bundler.write()`, and the early-exit callers (`?url`, `?inline`, html-linked) expand themselves. Before this, a `@directive;` in a stylesheet imported from JS was expanded inside rolldown's `load` hook while other modules were still being transformed, so it saw a partial registry (a 40-module repro alternated between 40 and 0 rules across runs). `e2e/native-plugin-build.mjs` now covers a JS-imported sheet with 24 marked modules.
- **Dev HMR: recompile before asking `invalidates`.** A changed module in the graph that a plugin's pass wants (or that a plugin already tracks) is recompiled first, so `module_seen` has run by the time `invalidates` is asked. A module that just gained the plugin's syntax is now picked up without the plugin reading the disk; the example plugin's workaround is gone. `e2e/native-plugin-dev.mjs` covers a plain module gaining a marker.
- **CI runs the example plugin's unit tests** (`cargo test -p oj --features example-plugin native_plugins`); `cargo test --workspace` never compiled them.

Left as follow-ups: TanStack Start dev and build do not receive native plugins yet; `config_section` only sees `oj.config.*` (vite-extract whitelists top-level keys); the process-global `NATIVE_PLUGINS` in build.rs; empty virtual sheets are still emitted; `mask` matches the exact directive spelling.

PR #174, opened 2026-09-10
fix(start): release live-reload sockets on client disconnect

TIMTOMTOP:
The TanStack Start live-reload handler waits only for rebuild notifications. When a client disconnects from `/@oj-start/hmr`, its socket and broadcast subscription stay in the handler until a later notification makes a write fail. Repeated connections can accumulate while the app is otherwise quiet.

Read from the socket alongside the reload channel, following the existing general HMR handler. This lets Axum process ping frames and lets the task exit on EOF or a socket error. On a close frame, finish the WebSocket close handshake before exiting so a pending reload cannot interrupt the acknowledgment. Reload messages and the editor's reload gate keep their existing behavior.

I reproduced this locally on main (`273f583`) before changing the loop:

- Real TCP WebSocket tests retained the reload subscription after both a close frame and an abrupt disconnect, without sending any reload notification.
- The real `oj dev` Start fixture timed out waiting for the close handshake.

Both pass with this change. The new socket tests also cover repeated connect/disconnect cycles, ping replies, ignored client text, and continued delivery of `reload` messages. A concurrent close/reload regression checks that the close acknowledgment preserves the client’s code and reason and releases the subscription; it fails without the explicit close handling. The existing Start HMR gate e2e now checks a clean disconnect before editing a file, then checks gated and immediate reloads as before.

Validation: `cargo test --workspace` and `node e2e/start-hmr-gate.mjs` pass. `cargo clippy -p oj --all-targets --no-deps` completes with warnings in unchanged code. `futures-util` supplies the socket close operation, and `tokio-tungstenite` is a test dependency. Both are already present in the workspace lockfile; no package versions change.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1826,6 +1826,7 @@ dependencies = [
  "blake3",
  "clap",
  "flate2",
+ "futures-util",
  "notify",
  "oj_cache",
  "oj_compiler",
@@ -1841,6 +1842,7 @@ dependencies = [
  "serde_json",
  "tokio",
  "tokio-stream",
+ "tokio-tungstenite",
 ]
 
 [[package]]
--- a/crates/oj/Cargo.toml
+++ b/crates/oj/Cargo.toml
@@ -31,8 +31,8 @@ oj_css.workspace = true
 axum.workspace = true
 notify.workspace = true
 tokio-stream = "0.1.19"
+futures-util = "0.3"
 
 [dev-dependencies]
 tokio = { version = "1.53.1", features = ["full", "test-util"] }
 tokio-tungstenite = "0.29"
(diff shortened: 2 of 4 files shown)


raphamorim:
Reviewed and verified — the leak and the missing close handshake are both real, and the fix + tests are solid (details in #183). #183 carries your commits unchanged (with authorship) plus one follow-up: the new loop broke the connection on a lagged broadcast receiver, where Vite's model (per-socket queue, nothing dropped) says a missed ping is just a pending reload — it now sends one reload and keeps the client, with a test. Only the branch changed — maintainer pushes to forks are blocked from this environment. Thanks for the excellent diagnosis and test coverage!

Issue #172, opened 2026-09-08
CSS sidecar truncates large JSON responses when stdin closes before stdout drains


L4Ph:
## Problem

`oj build` can fail on larger Tailwind stylesheets because the CSS sidecar exits before its piped stdout has drained. The child exits with status 0, but the JSON response is truncated. In my environment I captured exactly 65,536 bytes ending inside the `css` string.

The application-level error was:

```text
[UNLOADABLE_DEPENDENCY] Could not load src/index.css
plugin `oj:build` threw an error
```

## Environment

- oj 0.1.24, installed with `cargo install oj --version 0.1.24 --locked`
- macOS / arm64
- Node.js v24.20.0
- Original application: Tailwind 4.3.3
- Minimal reproduction: PostCSS 8.5.28, no React or Tailwind dependency needed
- Also reproduced directly with the sidecar from commit `273f583389dee1aa9c0f7db9615c58bcea1b7484`

## Minimal reproduction

Clone this repository, then create a separate scratch directory:

```sh
git clone https://github.com/raphamorim/oj.git
git -C oj checkout 273f583389dee1aa9c0f7db9615c58bcea1b7484
mkdir oj-sidecar-repro
cd oj-sidecar-repro
npm init -y
npm install postcss@8.5.28
printf 'export default { plugins: [] };
' > postcss.config.mjs
```

Save the following as `repro.py`. It supplies a valid large stylesheet, closes stdin as the build caller does, and briefly delays reading stdout to exercise backpressure. Python supplies actual pipes here; Node's `spawn` stdio can use sockets, which does not exercise the sidecar's `isFIFO()` branch.

```python
import json
from pathlib import Path
import subprocess
import sys
import time

base = Path.cwd()
css = '/*' + 'x' * (1024 * 1024) + '*/\n.a { color: red; }\n'
request = json.dumps({'id': 1, 'base': str(base), 'from': str(base / 'input.css'), 'css': css}) + '\n'
p = subprocess.Popen(['node', str(Path(sys.argv[1]).resolve())], stdin=subprocess.PIPE, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
p.stdin.write(request.encode())
p.stdin.close()
p.stdin = None
# Exercise stdout backpressure while the parent is briefly busy.
time.sleep(0.3)
out, err = p.communicate(timeout=10)
print({'exitCode': p.returncode, 'receivedBytes': len(out)})
try:
    response = json.loads(out)
[...]

PR #178, opened 2026-09-15
fix: flush Tailwind sidecar output before exit

saikotaa:
## Summary

Await Tailwind sidecar stdout writes before decrementing the in-flight count and exiting.

Without this, one-shot builds can terminate while Node still has buffered output. Large generated stylesheets are consistently truncated at the 64 KiB pipe boundary, causing `oj build` to fail while parsing the sidecar response.

## Verification

- `cargo test -p oj_server sidecar --lib`
- Reproduced against a Tailwind v4 application whose generated response exceeds 64 KiB; the build completes with this change.


raphamorim:
Thanks for the fix and the clear diagnosis — this was real (#172). It was superseded by #180, which merged first and covers all three compiler sidecars (tailwind, css-preprocess, svelte) plus a FIFO-transport regression test; the truncation repro from #172 is verified fixed on main. Closing in favor of #180.

PR #180, opened 2026-09-16
fix: wait for stdout to flush before exit

Brumor:
Fixes issue #172  by waiting for stdout to flush before exiting.


raphamorim (approved):
Verified end to end, approving:

- The new test fails on all three of main's sidecars and passes with the fix.
- Ran issue #172's exact Python repro against both: main truncates at 65,536 bytes (byte-for-byte the reported failure), this branch delivers the full 1,048,620-byte reply with valid JSON. This fixes the Tailwind `[UNLOADABLE_DEPENDENCY] Could not load src/index.css` build failure from #172.
- Full js unit suite on this branch: 0 failures.
- Scanned every other `process.exit(0)` in the assets: vite-extract emits its result to a file (its exit deliberately kills plugin-started watchers), ssr-runner/css-host/runner exit only on parent-death paths, tailwind `--once` exits naturally so writes drain. This PR covers exactly the three request/response-over-stdout sidecars — no gaps.

Nice test design: the shell FIFO pipeline (spawn's pipes can be sockets, which dodge the `isFIFO()` branch) plus the delayed reader for backpressure is exactly the right transport model.

Cosmetic only: `err? 1 : 0` spacing and dropped semicolons vs the surrounding style.

PR #198, opened 2026-09-18
Respawn the plugin host after a wedge

raphamorim:
Today `declare_gone` is terminal: once the in-process plugin host is declared gone (the transport belt's "the engine stopped scheduling", a failed boot, a dead push channel), every plugin-served transform fails "plugin host exited" until the dev server restarts. A single 40s wedge — a native block, a V8 GC storm, CPU starvation on a loaded machine — costs the whole session. The old process host effectively had supervisor restarts; the in-process host had nothing.


raphamorim:
Self-review round: three races fixed and amended in — (1) two racing revives could leave the losing ignite's engine alive (zombie isolate + middleware server); ignite now takes its expected generation and abandons its engine if superseded. (2) a stale push queued by a replaced engine (e.g. ojInit) could open the call gate before the new engine evaluated its module, racing the boot prelude; the dispatcher now drops everything from a superseded generation. (3) same hole via a hook reply from the old engine; the reply-driven initialized flip is now generation-guarded.

raphamorim:
Second self-review round over the full diff, two fixes amended in: (1) spawn_late_plugin_serve now clears its applied snapshot when it observes the revive's None reset, so a revived generation landing on the SAME middleware port still re-runs the catch-up resync (the code previously contradicted its own comment and would leave runner envs stale for dead-window edits); (2) shutdown() now serializes on the revive lock, closing a race where its engine-take could run before an in-flight ignite's install and leave a live zombie engine on a retired host — the install guard also checks the shut_down flag.

raphamorim:
Third review round, after main moved under the PR (#199 landed): #199 established that re-registering a native addon after every runtime that loaded it was destroyed is warned, not safe (pre-3.10 napi-rs addons SIGSEGV — vite 8.0.16's rolldown 1.0.3). The respawn created exactly that sequence on the clean-teardown path, converting a dead host into a possible whole-dev-server crash for that subset. Fixed: declare_gone snapshots the already-orphaned addons, and try_revive refuses (without burning budget, throttled by the spacing clock) exactly when an addon became orphaned by this host's own death — its successor runs the same plugins and would re-register it. Addons another engine still holds live (the usual case: Start/SSR engines share the toolchain) re-register as the ordinary concurrent-workers case and never block, and an orphan this host never loaded doesn't either. New oj_deno_napi query addons_pending_unsafe_reregistration() (re-exported through oj_js) backs the gate. Kept as a separate commit since it rides #199's registry.

raphamorim:
Research notes on the addon re-registration class the revive gate defends against (sources: napi-rs, Node, Deno, Bun, rolldown trackers):

**Root cause.** napi-rs keeps process-global state created at the FIRST registration and tied to the first isolate: the custom-GC threadsafe function (routes cross-thread Buffer/ref drops to "the" JS thread) and the env-cleanup/tokio-runtime gate. A process-wide "already initialized" latch means a registration into a fresh env after all envs died does not recreate them — the first custom-GC dispatch is then a use-after-free (napi-rs#3357, #2555).

**Fix boundary.** napi 3.10.0 routes custom-GC drops through the owning isolate (#3360); napi 3.11.0 registers the cleanup hook per registration (#3400) and is the fully-fixed floor. In rolldown/vite terms:

| vite | rolldown pin | napi-rs | status |
|---|---|---|---|
| 8.0.16 | 1.0.3 exact | 3.9.0 | affected (the #199 datapoint) |
| 8.1.0–8.1.3 | ~1.1.1–1.1.3 | 3.9.x | affected (tilde may float fixed) |
| 8.1.4–8.1.5 | ~1.1.4–1.1.5 | 3.10.x | custom-GC fixed |
| 8.2.1+ | ~1.2.1+ | 3.12.x | fully fixed |

**Prior art.** Node: init once per env, main thread never unloads, workers refcount+dlclose; NO guard against register-after-all-died — the addons doc puts process-global correctness on the addon, and Node segfaults the same way. Deno upstream (what we inherit) mem::forgets the library, caches the register fn per path, and re-runs it for every env with no live guard — same exposure as us, so the fork's registry + this gate is beyond-upstream hardening. Bun deliberately keeps napi envs alive at the end of `bun test` to dodge exactly this (bun#42746). Electron renderer reload is the flow napi-rs 3.10+ is designed for.

**Follow-up option (not this PR): keeper env.** Keeping one live env holding the addon converts the unsupported zero-live transition into the supported concurrent-workers case — but the keeper must be registered BEFORE the last holder dies (registering it after is itself the crash), and its event loop must keep pumping (custom-GC TSFNs dispatch to it). For the revive flow that means sequencing new-gen-registers-before-old-gen-unwinds, which is not generally controllable on a wedge — hence refuse-and-diagnose here, keeper-env as a designed follow-up if refused revives show up in the fleet. Caching the first exports across envs is unsound (V8 objects are single-context). Also noteworthy: rolldown 1.2.8's mimalloc 3.5.1 bump introduced a separate native-teardown SIGSEGV (rolldown#10860) — native teardown stays a moving target.

raphamorim:
Fourth self-review round, over the keeper commit; two refinements amended in: (1) the keeper now re-filters its addon set against live registrations at load time — an UNRELATED engine dying inside the keeper window orphans its addons, and the keeper requiring one of those would itself be the dangerous re-registration (the dying host's own addons always survive the filter: its env is held alive until the abandon that follows); (2) the push dispatcher drops pushes from a declared-dead generation, not just a superseded one — the keeper sequence keeps the dying engine alive briefly past its declaration, and a wedge that self-resolves in that window must not re-activate serve info for a middleware server the pending abandon is about to kill. Also audited clean: env-alive-until-abandon across all death shapes, lock ordering (revive → napi registry, keeper mutex leaf-only), double-declaration, keeper root staleness, and the test spacing-clear not reopening the crash window (a racing revive registers while the old env is still alive = the supported concurrent case). 176/176 across repeated runs, full e2e green.

PR #203, opened 2026-09-19
Allocation hygiene: what an arena-allocation review found (and did not)

williamhogman:
## Allocation hygiene: what an arena-allocation review actually found

This started as "implement arena allocation where appropriate". The workspace was surveyed for arena/bump/index-arena/interning opportunities, every candidate was adversarially challenged on bang-for-buck with a counting `GlobalAlloc` and release microbenches (scratch crates path-depending on the repo, repo untouched), the survivors were written up as proposals and judged, and only the ones that passed were implemented. The honest headline is that **true arena allocation does not pay off in oj**, and this PR lands the five allocation fixes the review did find instead.

### Arena verdict (measured, base 9d37c53)

Numbers in this section come from scratch harnesses (counting `GlobalAlloc`, release builds, path-deps on the repo crates) that were not added to the repo.

- **oxc `AllocatorPool` / `Allocator` reuse across `compile_module`**: saves ~4 mallocs of 55-277 per compile. `SemanticBuilder` allocates its own `Allocator` regardless of what the caller passes, and the compile only runs on a cache miss (the dev server returns from the memory / persistent cache first). Wall-clock within run-to-run noise.
- **`Allocator::with_capacity` pre-sizing** at the compile entry points: exactly 1 malloc saved per compile, and bytes requested from the system allocator went *up* on 3 of 5 real inputs (react-dom 14.7 -> 15.7 MiB). Dropped.
- **Index-arena `ModuleGraph`** (Vec + u32 ids, interned paths instead of `HashMap<PathBuf, ModuleNode>`): unmeasurable. The graph is touched for microseconds per request; the cost that did show up (52 allocations per warm `register_in_graph`) came from re-recording unchanged edges, not from the map's layout. Fixed by an early return instead (below).
- **mimalloc as global allocator**: within noise on the compile workload.
- Also refuted on hotness or cheaper alternatives: bump arenas for the CSS pipeline, index-keyed `pkg_bundle` graph, string interning of module ids, `__oj_rpc` bridge buffer reuse, code-cache single-buffer reads, a 32-byte cache key, and a parse-once fusion of the interop/CJS-detection passes (a correctness blocker: oxc lexes `<!--` differently in Module vs Unambiguous mode, so reusing the probe parse changes output for some `.js` files).

oxc and lightningcss already own the arena-shaped work. oj's remaining allocation cost was glue that copies or rebuilds data it already holds.

### What is changed (one commit each, plus a small follow-up tidy on item 5 and the CI fix in item 6)

1. **`crates/oj/src/build.rs`: transform hooks borrow `args.code` and return `Ok(None)` when nothing changed.** Lead item because it is a correctness fix. `OjCssPlugin::transform` returned `Some(expanded)` for every module containing `import(`, `import.meta.url` or `import.meta.glob` even when the expanders returned an identical copy. rolldown records that as a code change with an omitted map and collapses the module's sourcemap to `{"sources":[],"mappings":""}`; oj filters the `SOURCEMAP_BROKEN` warning, so with `build.sourcemap` on every `React.lazy` / route-split module silently lost its mappings. The hook body is now a pure `expand_import_meta(id, code) -> Option<String>` with a unit test and a production-build test that fails before and passes after (asserts `src/main.js` appears in its chunk's `.map`). Side effect: 8510 -> 4510 allocs on a 2000-module corpus; emitted JS identical.

2. **`crates/oj_css`: single-pass placeholder substitution.** After lightningcss prints with `analyze_dependencies`, each url()/@import placeholder was replaced with a whole-buffer `String::replace`: O(D*L) with 2D allocations. One quote-anchored pass (lightningcss always prints placeholders quoted, including inside `image-set()`) replaces both loops and is tested to match the old loop byte for byte on printed CSS. Substitution step, per call: 300 KB / 123 urls 6.8-19 ms and 142 MB requested -> 0.2-0.5 ms and 0.8 MB; end to end through `compile_css_dev` on the same sheet, 156.9 MB -> 15.9 MB requested (lightningcss itself is the rest). Component-sized CSS (D <= 10) will not notice; this is a tail fix for icon sets and data-URI-heavy generated CSS, mostly dev-side (`oj build` compiles with rebase=false).

3. **`crates/oj_server`: cached JS bodies are `axum::body::Bytes`.** `/@oj/chunk.js`, `/@oj/client.js`, `/@oj/bundle-runtime.js` and the `OJ_PARTIAL_BUNDLE` pkg caches copied the whole body into a fresh `String` per 200. `Bytes::from(String)` reuses the buffer, so a cache hit is a refcount bump; headers and wire bytes identical (a test pins the single `text/javascript`). 20 MB chunk.js per response: 20.97 MB / 338 us -> 839 B / 0.2 us. Under the full-reload HMR gate every frame refetches chunk.js as a 200 on every edit, so this is a memory-spike fix rather than a latency win.

4. **`crates/oj_graph`: skip re-recording unchanged edges.** `register_in_graph` runs on every dev module request (304s included) and per module in the bundle-mode rebuild loop, re-inserting present edges through allocating `ensure_module` / `add_import`. `set_imports` now takes `&[impl AsRef<Path>]`, returns early when the import set is unchanged (exact under repeats and reordering), and the lookups happen before any owned key is built. Warm call: 52.3 -> 3.3 allocations, ~-60% time; roughly 45 ms less serialized graph work per bundle-mode rebuild at 10k modules, low single-digit percent per dev request. The graph calls themselves (`set_imports` / `add_import` / `ensure_module` on a warm module) allocate nothing, pinned by a separate one-test binary with a per-thread counting allocator; the remaining ~3 per call are `register_in_graph`'s own `Vec` and lock. That test binary is deletable if it ever becomes a nuisance.

5. **`crates/oj_compiler`: cache the define list and its `ReplaceGlobalDefinesConfig` per (client, ssr) variant.** Every compile cloned the server-set define list (including the multi-KB `import.meta.env` blob) and every env-using module re-ran `ReplaceGlobalDefinesConfig::new`, an oxc parse per define value. The three setters now rebuild both variants eagerly; a compile is one `RwLock` read plus an `Arc` clone. Output fingerprints identical across 11 fallback/set/merge/re-set/ssr cases. ~50-pair env (40 `VITE_*` vars): env-using module 385 -> 87 allocs per compile, NODE_ENV dep 359 -> 61. New `tests/env_defines.rs`, and `examples/bench_compile --big-env` (49 pairs) reproduces the shape in-repo.

6. **`crates/oj` / `crates/oj_server`: the one-shot child reaper catches a parent that died before the child finished loading.** This is the fix for the `e2e (unbundled)` failure that blocked this PR and 9 of the last ~20 runs across branches. The `oj engine-job` / `oj start-script` ppid reaper compared against a snapshot taken once the child reached `main`; a parent SIGKILLed before that (a 1.6 GB debug binary on a slow CI disk takes over a second to get there) had already reparented the child, so the snapshot never changed and the orphan ran its whole job, writing ~135 V8 code-cache files into the app's `.oj-cache` for seconds after the dev server was gone: the recurring `ENOTEMPTY ... code-cache/v8-*` in the public-dir teardown. Every spawner now declares its pid in `OJ_PARENT_PID` and the reaper compares against that, so an already-reparented child exits on its first poll before an engine boots. Reproduced and A/B-verified on a Linux debug build with the parent killed 10 ms after spawn: main writes 0 -> 22 -> 103 -> 135 files and the result; the fix writes nothing and is gone before the first sample. The public-dir e2e also SIGKILLs the server's process group so its teardown does not depend on the child noticing, and a new unit test pins the declared-parent branch (the pre-change binary fails it).

### Deliberately not done

- Any pool, arena type, allocator swap or `unsafe` in shipped code. The one exception is the counting `#[global_allocator]` confined to the single-test binary `oj_graph/tests/warm_path.rs`. No new dependencies (`bytes` is axum's re-export).
- Plugin-hook RPC protocol trimming (transform reply as an object instead of a stringified string). Real (~57% less Rust heap per 20 KB transform in the scratch harness) but only exercised with a vite.config transform plugin, and the proposal wanted a counting allocator inside oj_server's existing shared test target rather than an isolated binary. Worth its own PR.
- `may_have_template_dynamic_import` prefilter in the transform hook (a further 4510 -> 3301 allocs in the scratch harness) and switching genuinely-changed modules to `HookTransformOutputMap::Null`: both are behavioural decisions that deserve their own tests.

### Verification

`cargo test --workspace` (all green, 0 failures), `cargo check --manifest-path fuzz/Cargo.toml --all-targets`, `cargo check -p oj_wasm --target wasm32-unknown-unknown`, `node --test e2e/unit/engine-child-orphan.test.mjs` (macOS and Linux), and `e2e/public-dir.mjs` 5/5 on a Linux debug build pinned to one contended core. The repo is not rustfmt-clean at HEAD, so `cargo fmt` was not run over whole files; new hunks follow each file's existing style. Playwright e2e was not run locally (not installed here); CI covers it.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


williamhogman:
**CI note: the `e2e (unbundled)` failure is a pre-existing flake, not this PR.**

The failing step is `publicDir files served verbatim` (`e2e/public-dir.mjs`), which dies in teardown with `ENOTEMPTY: directory not empty, rmdir '<app>/.oj-cache/v1/code-cache/v8-15.0.245.2-rusty'` about 5.5 s after the last check passes (the full `rmSync` retry budget), i.e. something keeps writing V8 code-cache files after the SIGKILLed server has exited.

Evidence it is independent of this branch:
- Control run #204 (main `9d37c53` + one empty commit, no code changes) fails the same step with the same error: https://github.com/lovablelabs/oj/pull/204
- The same failure hit `main` at `bf76abd` (run 35398699548) and 9 of the last ~20 `e2e (unbundled)` runs across branches on 2026-09-18, including four of six runs on the branch that landed #201's teardown fix.
- Locally on macOS the test passes 10/10 with both this branch's and main's debug `oj`; the race only reproduces on the Linux runner.
- Nothing in this diff touches the engine-job child, the reaper, the code cache, or the publicDir path; scaffold 2 (the one that fails) only fetches `/` and `/sw.js`.

The three reruns here failed the same way. I have not tried to fix the flake in this PR since it is unrelated to allocation work; #201's `SIGKILL + await exit + rmSync retries` clearly narrows but does not close the window, so it probably deserves its own follow-up.

williamhogman:
**Update: root cause found and fixed in `ddd354b`, superseding the "pre-existing flake" note above.**

The writer that survives the teardown is the boot-time dep-optimizer `oj engine-job` child. Its ppid reaper compared against a snapshot taken once the child reached `main`; when the dev server is SIGKILLed before the (1.6 GB debug) child binary has finished loading, the child is already reparented when it snapshots, so the reaper never fires and the child runs its whole job, writing ~135 code-cache files into the app dir for several seconds. That is the "bounded miss" the reaper's own comment described, and the slow CI runner makes it common.

Fix: every spawner declares its pid in `OJ_PARENT_PID`; the reaper compares against that and exits on its first poll. Reproduced and A/B-verified on a Linux debug build (parent killed 10 ms after spawn, six busy loops on the same core): main writes `0 -> 22 -> 103 -> 135` code-cache files plus the result; with the fix the child writes nothing. The public-dir e2e also kills the server's process group, and a new unit test pins the declared-parent branch (the pre-change binary fails it).

PR #206, opened 2026-09-21
Echo the oj-hmr identity subprotocol; let unclaimed upgrades dangle like Vite

raphamorim:


raphamorim:
CI red root-caused: the e2e teardown ENOTEMPTY was #201's documented born-orphan window materializing — the parent was SIGKILLed between spawning an extraction child and the child's ppid snapshot, so change-detection never fired and the cold-CI child out-wrote the rm retries. Closed for real: every child spawn announces its pid via OJ_PARENT_PID, and a child whose observed ppid differs at arm time exits immediately (deterministic born-orphan unit test added). public-dir 3/3, both orphan tests green, full battery green.

raphamorim:
Diagnostics named it: zero surviving oj processes and a complete, quiescent code-cache flush burst on disk — the writer is a child finishing naturally (~1s) right after the parent's SIGKILL, and `rmSync`'s `maxRetries` loses anyway because Node's sync rimraf only re-runs the failing rmdir on visited nodes; it never re-traverses, so a burst behind its traversal frontier defeats any retry count. Fix: the start.mjs pattern — retry the whole `rmSync` (fresh traversal per attempt, 20×150ms), which has kept start.mjs flake-free through the same class. Diagnostics stay in for anything that still survives fresh traversals.

PR #209, opened 2026-09-21
resolve bare imports inside node_modules with the Vite-style resolver

raphamorim:
Start SSR returned HTTP 500 before first render on any app whose dependency chain crosses an ESM→UMD seam: `@react-three/drei`'s ESM `core/DetectGPU.js` does `import { getGPUTier } from "detect-gpu"`, and detect-gpu's `main` is a minified UMD build with no `exports` map. typecheck and build pass while the dev server dies at module link with "The requested module 'detect-gpu' does not provide an export named 'getGPUTier'".


raphamorim:
Second commit from self-review: the Vite-style resolver has no builtin awareness, so the new seam would have resolved a dependency's bare `events`/`util`/`fs` import to an installed browserify-style polyfill package of the same name — where Node and Vite SSR (`fetchModule` checks `isBuiltin` before resolving) both give the builtin. Bare builtins now stay on Node semantics (`is_node_builtin` goes pub for the host), and the e2e grew a case: a junk `events` package squats on the name and the page must still render off the real EventEmitter.

Issue #188, opened 2026-09-17
darwin build is not bit-reproducible


raphamorim:
`nix build .#oj --rebuild` on aarch64-darwin reports `derivation may not be deterministic: output differs` for the same derivation built twice on the same machine.

Reproducible output would let nix-based consumers verify binaries by rebuild (2-of-2 builder agreement, independent republication checks) instead of trusting any single builder. Likely suspects are the usual mach-O nondeterminism sources: link member ordering, LC_UUID, ad-hoc codesignature timestamps. Linux reproducibility status unknown — worth checking with the same `--rebuild` probe.

First step: `nix build --rebuild --keep-failed` + `diffoscope` on the two outputs to localize the differing bytes.

raphamorim:
Progress — two distinct sources, one found and fixed:

**Source 1 (found, fix validated): embedded build-directory paths.** `strings` diff between two honest builds shows dependency panic-location paths under the nix build dir, which some nix implementations randomize per build (`/nix/var/nix/builds/nix-<pid>-<rand>/cargo-vendor-dir/...`). The two random names can even differ in length, shifting every subsequent byte — which is why ~65% of the binary appeared to differ. `RUSTFLAGS --remap-path-prefix $NIX_BUILD_TOP=/build` in the flake build makes the embedded paths build-dir-independent (verified: they now read `/build/cargo-vendor-dir/...`). Note this also means earlier LTO experiments were contaminated: fat/thin/off all looked nondeterministic because of the paths, not LTO.

**Source 2 (open): with paths remapped, the output still differs between rebuilds.** Byte-level localization of the remapped pair is in progress; candidates are ThinLTO parallel-backend ordering (rust-lang/rust#52044 class) or a dependency build script. Will follow up with the localization and either a fix or the measured cost of the deterministic configuration.

raphamorim:
Root-caused completely and fixed in #190. Source 2 was `aws-lc-sys`'s C sources embedding `__FILE__` build-dir paths (rustc's remap flag can't reach the C compiler); `-ffile-prefix-map` via `NIX_CFLAGS_COMPILE` covers it. With both remaps, `nix build --rebuild` is clean on aarch64-darwin with `lto = "thin"` and `codegen-units = 1` unchanged — ThinLTO is exonerated, it was 100% path embedding. No codegen or perf change.

PR #210, opened 2026-09-21
Compile from store paths so recorded extension paths are deterministic

raphamorim:
Since the Deno migration (#189), darwin builds of the flake are not bit-reproducible: two builds of the same derivation produce different binaries. This breaks any consumer that verifies reproducibility by building twice (#188 / #190 established that guarantee).

**Mechanism.** `deno_core`'s `extension!`/`include_js_files!` records every extension JS file by absolute path via `env!("CARGO_MANIFEST_DIR")` (deno_core 0.412 records paths unconditionally; the bytes come from the snapshot at runtime, but the path strings are always compiled in). `--remap-path-prefix` from #190 cannot rewrite them, because `env!` expands to a plain string literal before remapping applies. Modern Nix randomizes the build directory (`/nix/var/nix/builds/nix-<pid>-<rand>`), so 22 vendored `deno_*` crates plus `oj_deno_runtime` embedded ~20 different path strings per builder; ld64's cstring dedup/reordering then cascaded that into ~23MB of differing bytes.

**Fix: make the recorded paths independent of the build dir by compiling from store paths, which are identical on every builder of the same derivation:**
- `cargoDepsHook` points `cargoVendorDir` at a symlink to the vendor tree in the store (cargoSetupHook realpath-resolves it, skipping the build-dir copy), and `preBuild` rewrites `importCargoLock`'s relative `directory = "cargo-vendor-dir"` to the store path.
- The workspace builds from `$src` via `--manifest-path`, with `CARGO_TARGET_DIR` redirected to the writable build dir (kept cwd-relative for `cargoInstallHook`).
- `CARGO_HOME` is set to the hook-written config's directory so subprocesses that run from the store workspace (cargo-auditable's `cargo metadata`) still see the vendored-sources replacement.
- A `postInstall` guard greps the binary for `$NIX_BUILD_TOP` and fails the build on any leak, so this class of regression is caught by `nix.yml` CI on every PR.

Verified on aarch64-darwin: the binary now contains zero build-dir references (previously 20), all recorded extension paths are `/nix/store/...` paths, and `./result/bin/oj --version` works.

**Known remaining nondeterminism, out of scope here:** `deno_snapshots`' build script creates the V8 startup snapshot at build time, and that creation is not run-deterministic (~160KB of scattered serializer-ordering differences per run, reproduced twice in the same directory; unaffected by `--predictable --random-seed=42`, eager-parse flags, or a single-threaded V8 platform). With this PR, a repeated build differs *only* inside the embedded snapshot (plus the derived LC_UUID and code signature). Follow-up tracked in a separate issue.


raphamorim:
CI analysis of the first run:

- **Both linux nix builds**: the new leak guard fired — a false positive, fixed in 6ec0d7be. In the Linux sandbox `$NIX_BUILD_TOP` is the constant `/build`, which is also the `--remap-path-prefix` target, so a bare `grep "$NIX_BUILD_TOP"` matched legitimately remapped, deterministic strings (`/build/source/target/.../cranelift-codegen-*/out/...` from OUT_DIR paths in panic locations) and even innocent substrings like `src/build.rs`. When the build dir equals the remap target, raw and remapped paths are the same constant bytes — deterministic either way and indistinguishable — so the guard now only runs when the build dir is randomized (where any raw occurrence is a real leak), and prints the matching strings on failure.
- **unit tests**: one flaky assertion in `e2e/unit/cf-environment-api.test.mjs` ("the deciding source is logged") — it reads the plugin-host sidecar's captured stderr immediately after `serveInfo()` resolves and got an empty buffer. Unrelated to this PR (which only touches flake.nix; that job builds with plain cargo, and the same code passed CI 40 minutes earlier on #209's branch). Filed separately.
- **darwin nix build / e2e / prod smoke**: green on the first run.

raphamorim:
Review: the mechanism is right and, per research, the only one that can work — but as-is the PR ships a ~1.25 GiB closure regression. Details:

**Verified (all claims check out):**
- deno_core 0.412 records `concat!(env!("CARGO_MANIFEST_DIR"), ...)` unconditionally under the snapshot feature; `--remap-path-prefix` can't touch it (confirmed in the vendored source).
- The nixpkgs plumbing behaves exactly as described against the pinned rev: `cargoDepsHook` is eval'd first in `cargoSetupPostUnpackHook` (so setting `cargoVendorDir` skips the copy), the symlink is realpath-resolved, importCargoLock ships the relative `directory = "cargo-vendor-dir"` the sed rewrites, and the config lands at `$NIX_BUILD_TOP/.cargo/config.toml`. The `/build` skip in the guard is sound (Linux sandbox = constant dir).
- Ecosystem research says this approach is effectively the *only* fix for the `env!` class on darwin: no fixed-build-dir mechanism exists or is planned (nix#6520 open since 2022; Nix 2.30 randomized `/nix/var/nix/builds` deliberately), nixpkgs' remap direction (PR #561970) and cargo `trim-paths` (RFC 3127) both leave `CARGO_MANIFEST_DIR` unsanitized by design, and crane is direct prior art for the deps-from-store half (`[source.nix-sources] directory = /nix/store/...`). Compiling the workspace itself from `$src` has no prior art but nothing forbids it.

**Blocking finding — closure bloat.** Built the branch locally (aarch64-darwin): the embedded store-path strings make Nix's reference scanner promote build inputs to runtime dependencies.

- main-built oj: closure **173.1 MiB**, one direct reference (libiconv).
- this branch: closure **1.4 GiB**, five direct references: `cargo-vendor-dir` (1.2 GiB, transitively every vendored crate source), the `source` fileset, two `aws-lc-sys` crate paths, libiconv.

Every consumer (`nix copy`, the lovbox image, the darwin cache this PR exists to revive) would ship the extra ~1.25 GiB.

**Fix that composes with this PR:** the embedded paths are only *read* at snapshot-build time; in the installed binary they're dead strings. Post-process with `nuke-refs` (pkgs.nukeReferences) in `postFixup`, keeping the real link-time refs (`otool -L` on darwin / `patchelf --print-needed` + interpreter on Linux) via `-e`. nuke-refs rewrites the hash part to a constant `eeee…`, so the bytes stay identical across builders — reproducibility is preserved, the references are broken, closure returns to ~173 MiB. On aarch64-darwin the binary must be re-signed after mutation (stdenv's auto-sign fixup hook — needs ordering after the nuke, or an explicit `signDarwinBinariesIn`). The existing postInstall guard should then also assert the vendor/src hashes are gone, so the closure regression class is CI-caught like the path-leak class.

Happy to push that fixup to this branch if you want.

